Issue #145, opened 2012-10-08
Implement Tags API in Objects module


pengwynn:
Docs here: http://developer.github.com/v3/git/tags/

PR #155, opened 2012-10-14
Implement tag api. #145.

joeyw:
Adds .tag and .create_tag methods.

The create method uses a lot of parameters, would a hash be better suited for this? I went with required fields as params since thats what it feels like the rest of octokit uses.

Also the tags created with this seem to be invisible. I created a tag on a private repo and I can retrieve the tag with the sha I get when creating it, but it doesn't show up anywhere and can't be used for anything I can see. The tags don't show up on the GitHub repo page. Pulling the repo from GitHub doesn't mention the creation of any new tags either.

### DIFF
--- a/lib/octokit/client/objects.rb
+++ b/lib/octokit/client/objects.rb
@@ -75,6 +75,60 @@ module Octokit
         }
         post("repos/#{Repository.new(repo)}/git/blobs", options.merge(parameters), 3).sha
       end
+
+      # Get a tag
+      #
+      # @param repo [String, Hash, Repository] A GitHub repository.
+      # @param tag_sha [String] The SHA of the tag to fetch.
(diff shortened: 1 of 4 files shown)


ippeiukai:
Just sharing my experience:

I ended up using `Octokit::Client#post` directly because, although not documented, if tagger param is not present Github API seems to assume the tagger is the authenticated user with date now.

This is what I do now:
```Ruby
    # Octokit::Client#create_tag requires specifying tagger.
    # If tagger is not specified in the options, Github API automatically uses the authenticated user and current date.
    def octokit_client_create_tag2(repo, tag, message, object_sha, type, options = {})
      options.merge!(
        tag: tag,
        message: message,
        object: object_sha,
        type: type
      )
      octokit_client.post("#{Octokit::Repository.path(repo)}/git/tags", options)
    end
```

Issue #483, opened 2014-06-09
Look up repository information by ID


bryanbraun:
The API supports Github repository lookup by ID (here's an example: https://api.github.com/repositories/18749335).

In Octokit, the repository methods allow you to input a repo full_name ([example here](https://github.com/octokit/octokit.rb/blob/5323df945ecfd524556888e35d042a96c9055a1c/lib/octokit/client/repositories.rb#L12)) to get repository info.

It would be really convenient to lookup information by ID as well, because doing a call with a stored full_name can fail when somebody changes their username or repo name.

PR #485, opened 2014-06-15
Support id based repo endpoints

joeyw:
Add support for using repository IDs instead of `:owner/:name` for relevant methods.

``` ruby
Octokit.repo 417862
=> {:id=>417862, :name=>"octokit.rb"...}
```

cc: @bryanbraun Mind giving this a try?
#483


jm-janzen:
Thank you for this. I used it to link to the 1 billionth repo without spoiling the punchline: https://api.github.com/repositories/1000000000

PR #547, opened 2014-12-17
Make body argument optional on #create_issue and #create_pull_request

r7kamura:
## Background
GitHub has been adding more validations to the API v3. Recently, the validation logic for issue and pull request body was changed so that it doesn't allow `body` parameter with `null` value. For this API change, `create_issue(repo, title, nil)` and `create_pull_request(repo, base, head, title, nil)` begun to raise 422 error.

FYI: maybe related to https://github.com/octokit/octokit.rb/pull/538

## Changes
To fix this problem, I added the following changes in this Pull Request:

1. Pass a string value for the body, or no value at all
2. Make body argument optional


pengwynn:
Thanks for the patch! :cookie:

PR #553, opened 2015-01-23
Ensure Repository is valid when passed a string. Fixes #552

### LINKED ISSUE #552: Octokit::Client#repository with wrong input
`Octokit::Client#repository('test')` returns list of links from root url. Therefore `Octokit::Client#repository?('test')` returns `true` which is kind of wrong.
Shouldn't `Octokit::Client#repository('test')` throw `Octokit::NotFound` or `ArgumentError`?

joeyw:
Raise an error when a Repository object is initialized with a string
that is not valid user/repo format.

### DIFF
--- a/lib/octokit/repository.rb
+++ b/lib/octokit/repository.rb
@@ -18,6 +18,9 @@ module Octokit
         @id = repo
       when String
         @owner, @name = repo.split('/')
+        unless @owner && @name
+          raise ArgumentError, "Invalid Repository. Use user/repo format."
+        end
       when Repository
         @owner = repo.owner
         @name = repo.name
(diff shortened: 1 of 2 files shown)


joeyw:
A little simple, maybe not thought out. I haven't pushed any code in a short while and well, I just kinda needed to, ya know.

pengwynn:
Probably a good guard to have. I was playing around with a CLI thing the other night and found myself wanting to tweak this to handle both `owner/repo` and [several flavors of GitHub URL][urls].

[urls]: https://github.com/mislav/coral/blob/master/libexec/coral-filter-gh-project

pengwynn:
:moneybag:

PR #555, opened 2015-01-31
Update org spec to use ENV defined github login

sm11963:
Fixes test failures due to inconsistency between test cassette and spec.


pengwynn:
Thanks! :cake:

PR #557, opened 2015-02-03
Clearer instructions for list_issues

shkfnly:
These docs don't specify that list_issues only returns assigned issues by default


pengwynn:
Thanks for the patch, but I'd rather let the API docs cover the defaults so we don't get out of sync.

PR #559, opened 2015-02-09
Separate VERSION out into separate constants. Fixes #558.

### LINKED ISSUE #558: Exposing version parts as constants?
Hey,
One thing I have been liking in some gems I have created recently is this pattern of breaking out parts of the version string into constants (and then joining them for the full `VERSION`). Example:

```ruby
module Octokit
  MAJOR = 3
  MINOR = 7
  PATCH = 0
  VERSION = [MAJOR, MINOR, PATCH].join('.').freeze
end

>>  Octokit::MAJOR
# => 3

>>  Octokit::VERSION
# => "3.7.0"
```

I have been using this for upgrade/version detection and it's been working a treat as I have access to every part of the version without re-splitting it each time it's required.

Curious as to whether this would be something worth while adding to Octokit? I would be happy to open the pull request if you think it's a good idea.

jacobbednarz:


pengwynn:
Thanks

PR #562, opened 2015-02-13
Don't CGI escape label names

### LINKED ISSUE #560: remove_label fails with 404 when label name contains spaces
`client.remove_label(repo, pr, 'help wanted')` fails with a 404 response from the GitHub API even when the "help wanted" label exists, as the spaces in the label name aren't escaped correctly.

<pre>
2.0.0-p353 :002 > c.remove_label('domcleal/test', 2, 'help wanted')                                                                     
Octokit::NotFound: DELETE https://api.github.com/repos/domcleal/test/issues/2/labels/help+wanted: 404 - Label does not exist // See: https://developer.github.com/v3/issues/labels/#remove-a-label-from-an-issue
        from /home/dcleal/code/git/octokit.rb/lib/octokit/response/raise_error.rb:16:in `on_complete'
        from /home/dcleal/.rvm/gems/ruby-2.0.0-p353@tmp/gems/faraday-0.9.1/lib/faraday/response.rb:9:in `block in call'
        from /home/dcleal/.rvm/gems/ruby-2.0.0-p353@tmp/gems/faraday-0.9.1/lib/faraday/response.rb:57:in `on_complete'
        from /home/dcleal/.rvm/gems/ruby-2.0.0-p353@tmp/gems/faraday-0.9.1/lib/faraday/response.rb:8:in `call'
        from /home/dcleal/.rvm/gems/ruby-2.0.0-p353@tmp/gems/faraday-0.9.1/lib/faraday/rack_builder.rb:139:in `build_response'
        from /home/dcleal/.rvm/gems/ruby-2.0.0-p353@tmp/gems/faraday-0.9.1/lib/faraday/connection.rb:377:in `run_request'
        from /home/dcleal/.rvm/gems/ruby-2.0.0-p353@tmp/gems/faraday-0.9.1/lib/faraday/connection.rb:140:in `delete'
        from /home/dcleal/.rvm/gems/ruby-2.0.0-p353@tmp/gems/sawyer-0.6.0/lib/sawyer/agent.rb:94:in `call'
        from /home/dcleal/code/git/octokit.rb/lib/octokit/client.rb:339:in `request'
        from /home/dcleal/code/git/octokit.rb/lib/octokit/client.rb:171:in `delete'
        from /home/dcleal/code/git/octokit.rb/lib/octokit/client/labels.rb:88:in `remove_label'
        from (irb):2
</pre>

remove_label (and other methods in the same file) uses CGI.escape on the label name before making the request (https://github.com/octokit/octokit.rb/blob/568745f/lib/octokit/client/labels.rb#L88). CGI.escape is escaping the space as a + rather than %20, which the API doesn't appear to accept.

The request method escapes too (https://github.com/octokit/octokit.rb/blob/568745f/lib/octokit/client.rb#L339), but this isn't complete - it won't escape special URL characters such as "?" in label names, so removing CGI.escape would break those labels instead.  Unfortunately changing CGI.escape to another method (ERB::Util.url_encode?) that escapes spaces as %20 will only cause the % to be escaped again in the request method.

### LINKED ISSUE #561: Bug when handle labels
I can't update or delete a label with empty spaces or dash. But when I create its works.

Examples:
```ruby
Octokit.delete_label!("myorg/myrepo", "0 - BACKLOG")
Octokit.delete_label!("myorg/myrepo", "not planned")
```

PS: The given example for [delete a repository label](http://octokit.github.io/octokit.rb/Octokit/Client/Labels.html#delete_label%21-instance_method) isn't work.

pengwynn:
Properly handle labels with spaces. Remove `CGI.escape`, let Addressable handle it.

Fixes #560
Fixes #561


domcleal:
Thanks for the quick fix!

PR #567, opened 2015-02-20
Implement payload parser

zqzas:
@pengwynn Thank you :smiley:


lib/octokit/client/hooks.rb
```diff
@@ -277,6 +277,17 @@ def remove_org_hook(org, id, options = {})
         boolean_from_response :delete, "orgs/#{org}/hooks/#{id}", options
       end
 
+      # Parse payload string
+      #
+      # @param payload_string [String] The payload
+      # @return [Sawyer::Resource] The payload object
+      # @see https://github.com/octokit/octokit.rb/issues/453
+      def parse_payload(payload_string)
+        agent = Octokit.agent
```
pengwynn: I don't think we need this line since `agent` is already in scope here.

spec/octokit/client/hooks_spec.rb
```diff
@@ -131,4 +131,12 @@
       end
     end # .remove_org_hook
   end # with org hook
+
+  describe ".parse_payload", :vcr do
+    it "parses payload string" do
+      result = @client.parse_payload '{"action": "added", "scope": "team", "sender": {"site_admin": false}}'
```
pengwynn: Let's drop one of those fixtures in `spec/fixtures` and load it with the `fixture` helper just for consistency.

pengwynn:
Lookin' good, @zqzas.

pengwynn:
@kdaigle this might be relevant to your interests, too.

lib/octokit/client/hooks.rb
```diff
@@ -277,6 +277,16 @@ def remove_org_hook(org, id, options = {})
         boolean_from_response :delete, "orgs/#{org}/hooks/#{id}", options
       end
 
+      # Parse payload string
+      #
+      # @param payload_string [String] The payload
+      # @return [Sawyer::Resource] The payload object
+      # @see https://github.com/octokit/octokit.rb/issues/453
```
pengwynn: I think https://developer.github.com/v3/activity/events/types/ would make a good `@see` value here.

zqzas:
@pengwynn I've add fixtures into the spec. It reads a json from a fixture, parses it, and compares with the correct answer stored in another fixture.

spec/octokit/client/hooks_spec.rb
```diff
@@ -131,4 +131,13 @@
       end
     end # .remove_org_hook
   end # with org hook
+
+  describe ".parse_payload", :vcr do
+    it "parses payload string" do
+      test_json = fixture('create_payload.json').read
+      test_answer = fixture('create_payload_hash.txt').read
```
pengwynn: This approach might be overkill since the format of `inspect` could change if test dependencies are added to the project. I think we can just assert a couple of the fields at a couple of depths on `result` and be just fine.

zqzas: @pengwynn Yes! Code is updated :-)

pengwynn:
@zqzas Looks great. I'll see if @kdaigle has any input before I merge in the morning. :zap:

kdaigle:
@zqzas thanks so much for tackling this. I think this is a super sane way to parse payloads without having to deal with changing/updating webhook payloads in the future. Good stuff. :cake:

PR #570, opened 2015-03-06
Fix guard

### LINKED ISSUE #569: Error running specs via Guard
When I attempt to fire up guard, I get the following error:

```
$ ./script/guard
/opt/rubies/2.1.1/lib/ruby/gems/2.1.0/gems/notiffany-0.0.6/lib/notiffany/notifier/tmux.rb:199:in `rescue in _check_available': :tmux notifier is only available inside a TMux session. (Notiffany::Notifier::Base::UnavailableError)
...
```

Since I'm not running this inside a TMux session (:see_no_evil:), the tmux notifier ([configured in `Guardfile`](https://github.com/octokit/octokit.rb/blob/aa253271c7ce8eda9d4f6e7e82382a082805a886/Guardfile#L6-L11)) is not available.

Some other projects have run into a similar issue (e.g., https://github.com/guard/notiffany/issues/6). As a result, the [Guard wiki](https://github.com/guard/guard/wiki/Guardfile-DSL---Configuring-Guard#notification) encourages people to move this configuration out of a project's `Guardfile` and into individual users' dotfiles:

> NOTE: since notification is more of a user setting, it's best to have each team member set this in their own `~/.guard.rb` file (instead of the project `Guardfile`).

pengwynn:
Guard notifier settings are per-user and should be placed in `~.guard.rb`.

Fixes #569.

/cc @jasonrudolph

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ gem 'rake'
 
 group :development do
   gem 'awesome_print', :require => 'ap'
-  gem 'guard-rspec'
+  gem 'guard-rspec', '~> 4.5'
   gem 'hirb-unicode'
   gem 'pry'
   gem 'redcarpet'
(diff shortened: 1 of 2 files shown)


jasonrudolph:
:sparkling_heart:

PR #572, opened 2015-03-15
Implement licenses api

zqzas:
@pengwynn I've removed license_for_repository method.  :ok_hand:


lib/octokit/client/licenses.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      LICENSES_PREVIEW_MEDIA_TYPE = "application/vnd.github.drax-preview+json".freeze
+
+      # List all licenses
+      #
+      # @see https://developer.github.com/v3/licenses/#list-all-licenses
+      # @return [Array<Sawyer::Resource>] A list of licenses
+      # @example
+      #   Octokit.licenses
+      def licenses(license_name = '', options = {})
```
pengwynn: I don't think we need the `license_name` param here, right?

lib/octokit/client/licenses.rb
```diff
@@ -0,0 +1,50 @@
...
+        options = ensure_deployments_api_media_type(options)
+        get "licenses", options
```
pengwynn: Let's call `paginate` instead of `get` to futureproof this method.

lib/octokit/client/licenses.rb
zqzas: @pengwynn Check the update. There were only one method previously, and I decided to separate them but forget to delete that parameter....

lib/octokit/client/licenses.rb
```diff
@@ -0,0 +1,50 @@
...
+      LICENSES_PREVIEW_MEDIA_TYPE = "application/vnd.github.drax-preview+json".freeze
+
+      # List all licenses
+      #
+      # @see https://developer.github.com/v3/licenses/#list-all-licenses
+      # @return [Array<Sawyer::Resource>] A list of licenses
+      # @example
+      #   Octokit.licenses
+      def licenses(options = {})
+        options = ensure_deployments_api_media_type(options)
```
pengwynn: Ha, just noticed this should be `ensure_licenses_api_media_type`. :grinning:

zqzas:
@pengwynn :racehorse:

PR #576, opened 2015-04-15
Fix error not paginate with comment_comments. Auto paginate not working.

nguyen7thai:
I have problem when pull the commit_comments from github, the auto_paginate feature does not work. I checked in the commit_comments file and found out that, it used the method `get` instead of `paginate`.

### DIFF
--- a/lib/octokit/client/commit_comments.rb
+++ b/lib/octokit/client/commit_comments.rb
@@ -12,7 +12,7 @@ module Octokit
       # @return [Array] List of commit comments
       # @see https://developer.github.com/v3/repos/comments/#list-commit-comments-for-a-repository
       def list_commit_comments(repo, options = {})
-        get "#{Repository.path repo}/comments", options
+        paginate "#{Repository.path repo}/comments", options
       end
 
       # List comments for a single commit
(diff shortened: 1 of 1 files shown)


lib/octokit/client/commit_comments.rb
```diff
@@ -32,7 +32,7 @@ def commit_comments(repo, sha, options = {})
       # @return [Sawyer::Resource] Commit comment
       # @see https://developer.github.com/v3/repos/comments/#get-a-single-commit-comment
       def commit_comment(repo, id, options = {})
-        get "#{Repository.path repo}/comments/#{id}", options
+        paginate "#{Repository.path repo}/comments/#{id}", options
```
pengwynn: I don't think we need a change here since it's fetching a single commit comment.

pengwynn:
Thanks for the patch. One small change, and I think we can merge this.

lib/octokit/client/commit_comments.rb
nguyen7thai: Yes, you're right

pengwynn:
Thanks. :sparkling_heart:

PR #580, opened 2015-04-23
Add multi_json gem as test dependency

pengwynn:
Apparently required for VCR now. This should fix failing CI.

/cc @zqzas


zqzas:
@pengwynn :clap: 
My PR #579 has passed CI after adding `multi-json`.

Issue #211, opened 2013-02-02
Enterprise specific api support


joeyw:
- [ ] [Management console](https://support.enterprise.github.com/entries/23032152-management-console-api)
- [ ] [Admin stats](https://support.enterprise.github.com/entries/21188836-admin-stats-api)
- [ ] [Search indexing](https://support.enterprise.github.com/entries/23050336-search-indexing-api)

PR #573, opened 2015-03-25
Add support for user suspending and unsuspending

stetsenko:
Github API reference (requires Github Enterprise ):
 - https://developer.github.com/v3/users/administration/#suspend-a-user
 - https://developer.github.com/v3/users/administration/#unsuspend-a-user

PR #582, opened 2015-04-27
Add support for the Enterprise admin APIs

gjtorikian:
In late-2013 (:scream_cat:), @joeyw opened https://github.com/octokit/octokit.rb/pull/388 to support [the GitHub Enterprise admin APIs](https://developer.github.com/v3/enterprise/). /cc #211 #306

In mid-2014 (:scream:) I started picking up on his work to continue implementing the API. By the end of 2014, both of our work was somewhat obsolete, as [Enterprise 2.0](https://enterprise.github.com/releases/2.0.0) was released and changed the game.

This PR reimagines the Enterprise API work by creating two separate clients: `EnterpriseAdminClient` and `EnterpriseManagementConsoleClient `. The former accesses the admin-only APIs, while the latter accesses just the management console APIs, which require a different API endpoint and a different authentication scheme.

The admin API endpoints supported in this PR are:

* [Admin Stats](https://developer.github.com/v3/enterprise/admin_stats/)
* [License](https://developer.github.com/v3/enterprise/license/)
* [Search Indexing](https://developer.github.com/v3/enterprise/search_indexing/)

Other APIs ([user administration](https://developer.github.com/v3/users/administration/), [LDAP](https://developer.github.com/v3/enterprise/ldap/)) should be handled in subsequent PRs, because this one is quite large as it is.


stetsenko:
@pengwynn thanks a lot, really excited for the official Enterprise API support. It will come really handy.
Is user administration coming next or should I port https://github.com/octokit/octokit.rb/pull/573 to EnterpriseAdminClient (or EnterpriseUserClient?).

pengwynn:
@stetsenko If you could port #573 to EnterpriseAdminClient that would be awesome. :sparkles:

lib/octokit/client.rb
```diff
@@ -325,6 +342,14 @@ def octokit_warn(*message)
 
     private
 
+    def endpoint
+      if self.class.to_s =~ /EnterpriseManagementConsoleClient/
```
pengwynn: @gjtorikian Can we use `self.instance_of?(EnterpriseManagementConsoleClient)`?

lib/octokit/client/pub_sub_hubbub.rb
```diff
@@ -93,7 +93,7 @@ def unsubscribe_service_hook(repo, service_name)
       private
 
       def pub_sub_hubbub_request(options = {})
-        # This method is janky, bypass normal stack so we don'tl
+        # This method is janky, bypass normal stack so we don't
```
pengwynn: :heart:

lib/octokit/client.rb
pengwynn: Better yet, can we just mark this method as `protected` and let that type override?

stetsenko:
@pengwynn cool, will do!

lib/octokit/client.rb
gjtorikian: > Better yet, can we just mark this method as protected and let that type override?

Nice, way better way to handle this. :sparkling_heart:

gjtorikian:
@pengwynn very rightly noted that it was possible for the `EnterpriseManagementConsoleClient` to call the DotCom client methods, because *client.rb* (from which everything inherits) also `include`s every `Octokit::Client::*` module. The idea of a "client" is right now mixed up between "a base set of methods from which everything inherits" and "the client to access the DotCom APIs."

There's two ways to approach this, I think. I took the less messy one in 58ab441. We'll only include those DotCom modules if the client is not an instance of one of the new Enterprise clients.

The messier way would be to split out a new class, perhaps called `DotCom`, so that we include `Octokit::DotCom::*` for just the dotcom client. I didn't think too hard about it, but it feels like this would be a breaking change, as suddenly existing clients would have an `is_a?` of `Octokit::DotCom`, not `Octokit::Client`. I don't know if that's a concern or not, or what other errors could creep in. This would be The Right Thing to Do&trade;, however.

gjtorikian:
@pengwynn A gentle bump. Any thoughts on the write-up above and the corresponding commit to fix faulty inheritance?

pengwynn:
@gjtorikian The latest change feels a bit brittle. I'm leaning toward keeping `Octokit::Client` as the dotcom client and extracting its shared plumbing into `Octokit::BaseClient` or something. Let me take a deeper look and ping you :soon:.

stetsenko:
@pengwynn FYI I have the changes ready for migrating #573 (and also user promote and demote) to EnterpriseAdminClient: https://github.com/octokit/octokit.rb/compare/enterprise-admin-api...stetsenko:enterprise-user-api.

Once this PR is merged, I can create a follow up one. Unless you think these changes should go into this branch?

Thanks and looking forward to having Enterprise API support in octokit! :telescope:

pengwynn:
@stetsenko Thanks! I'm going to fold in the changes from https://github.com/octokit/octokit.rb/compare/enterprise-admin-api...extract-connection shortly and merge this PR. I'll cherry pick your changes from there. :sparkling_heart:

fotinakis:
It looks like this PR caused a full regression of this change: https://github.com/octokit/octokit.rb/pull/583 — the `paginate` method in the new connection.rb no longer passes headers through to followup requests, so pagination is broken for things like the new organizations API media type.

gjtorikian:
Indeed. Looks like `master` might not have been merged in before the method split occurred. @pengwynn I'll open a PR shortly.

lib/octokit/enterprise_admin_client/admin_stats.rb
```diff
@@ -95,7 +95,7 @@ def admin_milestones_stats
       def admin_gists_stats
         get_admin_stats "gists"
       end
-      
+
       # Get only comment-related stats
```
myapcom: "���/E����|I���g��
�:tO�ﻍX!g&�i8os��*Z�x�w��P/��#����B|
�z���h[�Q=���
"���/E����|I���g��
�:tO�ﻍX!g&�i8os��*Z�x�w��P/��#����Bn<�}k��9ٻARw���
"���/E����|I���f��
�:tO�ﻍX!g&�i8os��*Z�x�w��P/��#����Bn7�St��ɝ2v��FB����
"���/E����|I���e��
�:tO�ﻍX!g&�i8os��*Z�x�3@��D2��Y���F�Msz��1 X�����
"���/E����|I���e��
�:tO�ﻍX!g�Cdf��*Z�x�u�����
"���/E���
�~O���a��
�;tN�⻍T!g!�Q4og��kk��`��D��z���OKz�8rhϧ�I٨X[�x���v�-D�����
"���/E���
�~O���j��
�;tN�⻍X!g$�qj9�ͻ=}�E�0V��X(��}���W_���
"���/E���
�~O���b��
�;tN�⻍X!g�Cdf��*Z�x�r�����
"���/E���
�~O���a��
�;tN�⻍X!g�Cdf��*Z�x�u�����
"���/E���
�~O���`��
�;tN�⻍X!g&�i8os��*Z�x�w��P/��#����B|
�qت��3Vެ>���
"���/E���
�~O���g��
�;tN�⻍X!g&�i8os��*Z�x�w��P/��#����B|
�z���h[�Q=���
"���/E���
�~O���f��
�;tN�⻍X!g&�i8os��*Z�x�w��P/��#����Bn<�}k��9ٻARw���
"���/E���
�~O���e��
�;tN�⻍X!g&�i8os��*Z�x�w��P/��#����Bn7�St��ɝ2v

PR #588, opened 2015-05-13
#edit_hook should not have defaults

byroot:
Because of theses defaults `edit_hook` is basically unusable. Imagine you want to add an event type:

```
client.edit_hook('my/repo', 'web', {}, add_events: ['status'])
```

This will actually mutate your hook to offer only the `push` event.

@pengwynn for review please.


pengwynn:
> This will actually mutate your hook to offer only the push event.

Or make an inactive hook live. :dizzy_face: Thanks for the patch. :cake:

kdaigle:
:heart:

Issue #586, opened 2015-05-11
Rate limit resource


ivantsepp:
https://developer.github.com/v3/rate_limit/ endpoint returns information for both core and search API requests. Is there a way to surface the number of search requests you have remaining? I think Octokit has information about core requests (and it looks like it's retrieved from the response headers). 

If Octokit doesn't support this currently, would it be acceptable to change it so that it does?

pengwynn:
Yeah, that resource was expanded after the rate limit helpers were written for Octokit. We may have to ditch the way we read the values from the `last_response` since that last response could be a "core" or "search" resource. Would definitely entertain a patch if you want to open a PR for discussion.

ivantsepp:
Yup! I'll open a pull request one soon, thanks!

blunderdome:
I'm not sure why the referenced PR never made it through, but I would also love to be able to get the search limit remaining.

kytrinyx:
@blunderdome @ivantsepp -- it looks like it was closed due to drift in the codebase. If either of you would be interested in carrying the pull request (see section 4 in https://blog.jessfraz.com/post/the-art-of-closing/) / reimplementing the changes against a fresh master, that would be superb.

PR #589, opened 2015-05-18
Change rate_limit to include search information

ivantsepp:
Sorry for the very late response!

See #586. This PR is to start a discussion on the various different paths one could take to tackle #586.

This PR:
 1. Changes `rate_limit` in that it will return information from the last time you called `rate_limit!` instead of getting the information from the last response.
 2. Changes `rate_limit!` in that it returns a structure that contains information about search and core resources

Notes:
For item 1, I can see a way we can maintain the previous goal of not hitting `/rate_limit` while still containing the latest info. We could store references to the previous "core" response and previous "search" response and adjust accordingly.


lib/octokit/client.rb
```diff
@@ -338,7 +338,8 @@ def request(method, path, data, options = {})
         end
       end
 
-      @last_response = response = agent.call(method, URI::Parser.new.escape(path.to_s), data, options)
+      response = agent.call(method, URI::Parser.new.escape(path.to_s), data, options)
+      @last_response = response if !path.include?('rate_limit')
```
ivantsepp: This is probably a bad way to detect this. Maybe we should do `!path.start_with('/rate_limit')`?

pengwynn:
Gonna close this out because of drift, but we'll look at what we can do for #586.

PR #591, opened 2015-05-22
Follow repository redirects

pengwynn:
Cherry-picking work done by @jasonrudolph, this PR adds support for [Repository redirects][blog] when passing the preview media type.

```
Octokit.repo "pengwynn/octokit"
=> Octokit::NotFound: GET https://api.github.com/repos/pengwynn/octokit: 404 - Not Found // See: https://developer.github.com/v3

Octokit.repo "pengwynn/octokit", :accept => "application/vnd.github.quicksilver-preview+json"
=> {:id=>417862, :name=>"octokit.rb", :full_name=>"octokit/octokit.rb", ...
```

[blog]: https://developer.github.com/changes/2015-04-17-preview-repository-redirects/


jasonrudolph:
@kdaigle and I sketched out this plaintext specification representing our recommendations for GitHub API clients following redirects:

> * **MUST** follow redirect requests if HTTP status code is 301, 302, or 307. (These are the redirect codes actively used by the GitHub API.)
> * **MUST** redirect a 30x status code if the HTTP method is HEAD, OPTIONS, GET, POST, PUT, PATCH, or DELETE.
> * **MUST** use the original request's HTTP method when following a redirect where the HTTP status code is 307.
> * **SHOULD** use the original request's HTTP method when following a redirect where the HTTP status code is 301 or 302.
> * **MUST** use the original request's authentication credentials when following a redirect where the original host matches the redirect host.
> * **MUST NOT** use the original request's authentication credentials when following a redirect where the original host does not match the redirect host.
> * **SHOULD** only follow 3 redirects.

@pengwynn: As part of the review for this pull request, we should verify that we satisfy these recommendations. For example, I don't think this pull request satisfies this recommendation yet (and their might also be others we're missing):

> * **MUST NOT** use the original request's authentication credentials when following a redirect where the original host does not match the redirect host.

pengwynn:
@jasonrudolph: What do you think of d97eab7...fe007ad?

btown50569:
Aaron

spec/octokit/client_spec.rb
```diff
@@ -422,6 +422,134 @@
...
+    end
+
+    it "does not change HTTP method when following a redirect" do
+      client = oauth_client
+
+      original_request = stub_get("/foo").
+        to_return(:status => 301, :headers => { "Location" => "/bar" })
+      redirect_request = stub_get("/bar").to_return(:status => 200)
+
+      client.get("/foo")
```
jasonrudolph: If an HTTP client was going to change the verb when following the redirect, it's most likely to change the verb to a `GET`. To increase this test's ability to catch future regressions, I recommend using something other than a `GET` for this test (e.g., maybe use a `DELETE`).

jasonrudolph:
> What do you think of d97eab7...fe007ad?

:star2: 

I noted one small suggestion in https://github.com/octokit/octokit.rb/pull/591#discussion_r31930550. Overall, this looks good to go. :rocket:

lib/octokit/middleware/follow_redirects.rb
```diff
@@ -0,0 +1,131 @@
...
+            response = perform_with_redirection(new_request_env, follows - 1)
+          end
+        end
+        response
+      end
+
+      def update_env(env, request_body, response)
+        original_url = env[:url]
+        env[:url] += safe_escape(response["location"])
+        unless same_host?(original_url, env[:url])
```
manojmj92: Hello, can you point me to examples where GitHub redirects you to URLs on a different host? I could not find examples online and I am wondering why this line has been added?

PR #596, opened 2015-06-07
Paginate fetch organizations

arunagw:
Found that organizations won't show up if there are organizations more
than 30. This will allow use of `auto_paginate` and fetch all
organizations.


pengwynn:
:cake:

arunagw:
great! thanks.

PR #593, opened 2015-05-24
Stop appending headers to querystring

krainboltgreene:
Fixes #592

PR #597, opened 2015-06-08
Incorporate connection changes that were made

gjtorikian:
@fotinakis rightly noticed that [some changes made to *client.rb*](https://github.com/octokit/octokit.rb/pull/582#issuecomment-110148506) were lost to the sands of time. This comes from 82ea118ec525b30a3d768c0cd2d04b4c9824dd62, which saw a bunch of shareable client connection methods moved out into their own module.

I went through the Git history and found two PRs that needed to be brought back in:

* https://github.com/octokit/octokit.rb/pull/583
* https://github.com/octokit/octokit.rb/pull/593

For an extra sanity check, it would be great if someone could verify that these are indeed the only client changes.

/cc @pengwynn for :eyes:

### DIFF
--- a/lib/octokit/connection.rb
+++ b/lib/octokit/connection.rb
@@ -81,11 +81,11 @@ module Octokit
         opts[:query][:per_page] ||=  @per_page || (@auto_paginate ? 100 : nil)
       end
 
-      data = request(:get, url, opts)
+      data = request(:get, url, opts.dup)
 
       if @auto_paginate
         while @last_response.rels[:next] && rate_limit.remaining > 0
-          @last_response = @last_response.rels[:next].get
+          @last_response = @last_response.rels[:next].get(:headers => opts[:headers])
           if block_given?
             yield(data, @last_response)
           else
--- a/lib/octokit/connection.rb
+++ b/lib/octokit/connection.rb
@@ -179,7 +179,7 @@ module Octokit
     end
 
     def parse_query_and_convenience_headers(options)
-      headers = options.fetch(:headers, {})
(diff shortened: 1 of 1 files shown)


fotinakis:
There were also tests added in https://github.com/octokit/octokit.rb/pull/583/files that don't seem to be included here.

gjtorikian:
The tests [are already in `master`](https://github.com/octokit/octokit.rb/blob/4e0f5ad48aeccc945011b9eb861566c51cfd868e/spec/octokit/client_spec.rb#L585-L591), because Git.

fotinakis:
Ah gotcha. I assumed they weren't because things are passing, they didn't catch the merge failure?

pengwynn:
>  I assumed they weren't because things are passing, they didn't catch the merge failure?

Looks like VCR masked it. Removing the cassette produces a failure:

```
Failures:

  1) Octokit::Client auto pagination sets headers for all pages
     Failure/Error: expect(Octokit.client.last_response.headers["x-github-media-type"]).to eq(github_media_type_header)

       expected: "github.v3; param=text-match; format=json"
            got: "github.v3; format=json"

       (compared using ==)
     # ./spec/octokit/client_spec.rb:590:in `block (3 levels) in <top (required)>'

Finished in 18.82 seconds (files took 0.46401 seconds to load)
543 examples, 1 failure
```

Thanks for the patch @gjtorikian. I'll give this a thorough once-over and merge tomorrow.

pengwynn:
Cool, I blew away the cassettes and this branch passes. Thanks! :cake:

PR #615, opened 2015-06-25
Handle invalid argument types passed to Repository

### LINKED ISSUE #614: Passing Octokit.repository? bad arguments returns true
```
2.0.0-p353 :001 > Octokit.repository?(nil)
 => true 
2.0.0-p353 :002 > Octokit.repository?(false)
 => true 
2.0.0-p353 :003 > Octokit.repository?(true)
 => true 
```

It should return `ArgumentError: Invalid Repository. Use user/repo format.` like it does if you pass `Octokit.repository?("foo")`.

henare:
Fixes #614.


pengwynn:
Thanks. :cake:

PR #620, opened 2015-06-29
Allow creating repo under org using organization ID instead of owner/name

pengwynn:
Builds on #619, adding a failing test to be sure future :penguin: doesn't break this. 

/cc @tarebyte @kdaigle.

Closes #619.

### DIFF
--- a/lib/octokit/client/repositories.rb
+++ b/lib/octokit/client/repositories.rb
@@ -155,7 +155,7 @@ module Octokit
         if organization.nil?
           post 'user/repos', options
         else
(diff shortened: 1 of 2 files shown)


tarebyte:
@pengwynn :sparkling_heart:

tarebyte:
@pengwynn is there a chance you can release new version so that I can use this? If not I'm cool with just pointing to master for now.

pengwynn:
@tarebyte I just released v4.0.1. Thanks for the fix.

PR #621, opened 2015-07-02
Add #pages_endpoint to Octokit::Default

parkr:
Helps ease GitHub Pages switch from dotcom to enterprise

### DIFF
--- a/lib/octokit/default.rb
+++ b/lib/octokit/default.rb
@@ -19,6 +19,9 @@ module Octokit
 
     # Default WEB endpoint
     WEB_ENDPOINT = "https://github.com".freeze
+    
+    # Default GitHub Pages endpoint
+    PAGES_ENDPOINT = "https://github.io".freeze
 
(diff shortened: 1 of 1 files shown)


parkr:
Not sure about the naming.

/cc @pengwynn

pengwynn:
Hey @parkr. I'm not clear on how this helps. Where is it consumed?

parkr:
We use it in [github-metadata](https://github.com/jekyll/github-metadata/blob/44f54079585badd4101f7db6a948fbd7fd98e065/lib/jekyll-github-metadata/pages.rb#L18) gem for Pages's `site.github` namespace plugin. It's used to [determine the user's Pages URL](https://github.com/jekyll/github-metadata/blob/44f54079585badd4101f7db6a948fbd7fd98e065/lib/jekyll-github-metadata/repository.rb#L87-L116), e.g. `pengwynn.github.io` or `pages.github.com`.

It would be most useful if there were a way to encapsulate this check in Octokit, as anyone using this library with either Enterprise or dotcom could reliably determine Pages URL's for a given account and/or repo.

PR #625, opened 2015-07-31
Suggest installing the newly-released 4.x gem

nilbus:
…unless you think it's not ready for prime time yet. ;-)


gjtorikian:
Thanks!

PR #627, opened 2015-08-04
Migrator support

gjtorikian:
This PR:

* Implements the new Migrations API bits and pieces 
* Abstracts the "preview header" logic so that it can be used by multiple endpoints

/cc @jonmagic @spraints @pengwynn


lib/octokit/client/organizations.rb
```diff
@@ -595,6 +595,89 @@ def remove_organization_membership(org, options = {})
...
+      #
+      # @param org [String, Integer] Organization GitHub login or id.
+      # @param id [Integer] ID number of the migration.
+      # @see https://developer.github.com/v3/orgs/migrations/#download-a-migration-archive
+      def download_migration_archive(org, id, options = {})
+        options = ensure_api_media_type(:migrations, options)
+        url = "orgs/#{org}/migrations/#{id}/archive"
+
+        response = client_without_redirects(options).get(url)
+        response.body
```
spraints: Should this be the "Location" header from the response?

lib/octokit/client/organizations.rb
```diff
@@ -595,6 +595,89 @@ def remove_organization_membership(org, options = {})
         user = options.delete(:user)
         user && boolean_from_response(:delete, "orgs/#{org}/memberships/#{user}", options)
       end
+
+      # Initiates the generation of a migration archive.
+      #
+      # Requires authenticated organization owner.
+      #
+      # @param org [String, Integer] Organization GitHub login or id.
+      # @option options [Array<String>] :repositories Repositories for the organization.
```
spraints: The list of repositories is required by the API. Would it make sense to move the list of repositories to a normal argument (`def start_migration(org, repositories, options = {})`)?

spraints:
:sparkling_heart: Thanks for adding support for the new API @gjtorikian!

lib/octokit/client/organizations.rb
gjtorikian: I just checked, and you're right, `Location` provides the same information as the body:

```
   body=
    "https://s3.amazonaws.com/github-cloud/migration/102/89?response-content-disposition=filename%3Db8902e90-3c6b-11e5-8f00-4be3b3f48d46.tar.gz&response-content-type=application/x-gzip&AWSAccessKeyId=...",
...
   response_headers=
    {"server"=>"GitHub.com",
     "date"=>"Thu, 06 Aug 2015 18:50:12 GMT",
...
     "location"=>
      "https://s3.amazonaws.com/github-cloud/migration/102/89?response-content-disposition=filename%3Db8902e90-3c6b-11e5-8f00-4be3b3f48d46.tar.gz&response-content-type=application/x-gzip&AWSAccessKeyId=...",
     "x-xss-protection"=>"1; mode=block",
```

Do you have a preference in terms of longer term stability?

lib/octokit/client/organizations.rb
gjtorikian: :+1: Sounds great.

gjtorikian: Do you think `repositories` should be a full list of NWO or just the names of repositories? 

I could do `repositories.map { |short_repo_name| "#{org}/#{short_repo_name}" }` if the latter. 

/cc @pengwynn for design consideration.

jonmagic: Either form works so I don't have a preference.

lib/octokit/client/organizations.rb
spraints: > Do you have a preference in terms of longer term stability?

I choose ... the `Location` header!

lib/octokit/client/organizations.rb
gjtorikian: You know what, I'm going to leave it as NWO. This is how the `.create_team` method prefers it too: https://github.com/octokit/octokit.rb/blob/1e8b395ba14a6d5d25b7dbe82a6a22a42ce4c585/lib/octokit/client/organizations.rb#L241-L246

Will definitely move it in as a second parameter though. `.create_team` can leave it in the options because it's not required. (I suspect I just stole the idea from this method entirely.)

lib/octokit/client/organizations.rb
```diff
@@ -595,6 +595,87 @@ def remove_organization_membership(org, options = {})
...
+      end
+
+      # Fetches the URL to a migration archive.
+      #
+      # Requires authenticated organization owner.
+      #
+      # @param org [String, Integer] Organization GitHub login or id.
+      # @param id [Integer] ID number of the migration.
+      # @see https://developer.github.com/v3/orgs/migrations/#download-a-migration-archive
+      def download_migration_archive(org, id, options = {})
```
pengwynn: I know the docs call the method this, but in Octokit land I think this should probably be `migration_archive_url` since the caller will have to do something with the result. Octokit won't do the download.

gjtorikian: Good call.

pvdb:
There's a minor problem with the refactoring in this PR, especially with the changes in commit 1e2bba0: the message issued by the `Octokit::Preview.warn_preview` method is now always warning about the "Migration API", irrespective of the preview type passed into the `Octokit::Preview.ensure_api_media_type` method, as illustrated here:

```ruby
[9] pry(main)> Octokit::Preview.send(:module_function, :ensure_api_media_type)
=> Octokit::Preview
[10] pry(main)> Octokit::Preview.ensure_api_media_type(:migrations, {})
WARNING: The preview version of the Migrations API is not yet suitable for production use.
You can avoid this message by supplying an appropriate media type in the 'Accept' request
header.
=> {:accept=>"application/vnd.github.wyandotte-preview+json"}
[11] pry(main)> Octokit::Preview.ensure_api_media_type(:licenses, {})
WARNING: The preview version of the Migrations API is not yet suitable for production use.
You can avoid this message by supplying an appropriate media type in the 'Accept' request
header.
=> {:accept=>"application/vnd.github.drax-preview+json"}
[12] pry(main)> Octokit::Preview.ensure_api_media_type(:blegga, {})
WARNING: The preview version of the Migrations API is not yet suitable for production use.
You can avoid this message by supplying an appropriate media type in the 'Accept' request
header.
=> {:accept=>nil}
[13] pry(main)> _
```

The problem is that the method is invoked as:

```ruby
warn_preview(:migrations)
```

... instead of:

```ruby
warn_preview(type)
```

I will issue a PR to fix this oversight shortly.

spec/cassettes/Octokit_Client_Organizations/_migrations/starts_a_migration_for_an_organization.json
```diff
@@ -0,0 +1 @@
...h1Yi5jb20vPEdJVEhVQl9URVNU\nX09SR0FOSVpBVElPTj4vYXBpLXBsYXlncm91bmQuZ2l0Iiwic3NoX3VybCI6\nImdpdEBnaXRodWIuY29tOjxHSVRIVUJfVEVTVF9PUkdBTklaQVRJT04+L2Fw\naS1wbGF5Z3JvdW5kLmdpdCIsImNsb25lX3VybCI6Imh0dHBzOi8vZ2l0aHVi\nLmNvbS88R0lUSFVCX1RFU1RfT1JHQU5JWkFUSU9OPi9hcGktcGxheWdyb3Vu\nZC5naXQiLCJzdm5fdXJsIjoiaHR0cHM6Ly9naXRodWIuY29tLzxHSVRIVUJf\nVEVTVF9PUkdBTklaQVRJT04+L2FwaS1wbGF5Z3JvdW5kIiwiaG9tZXBhZ2Ui\nOm51bGwsInNpemUiOjExMiwic3RhcmdhemVyc19jb3VudCI6MCwid2F0Y2hl\ncnNfY291bnQiOjAsImxhbmd1YWdlIjpudWxsLCJoYXNfaXNzdWVzIjp0cnVl\nLCJoYXNfZG93bmxvYWRzIjp0cnVlLCJoYXNfd2lraSI6dHJ1ZSwiaGFzX3Bh\nZ2VzIjpmYWxzZSwiZm9ya3NfY291bnQiOjAsIm1pcnJvcl91cmwiOm51bGws\nIm9wZW5faXNzdWVzX2NvdW50IjowLCJmb3JrcyI6MCwib3Blbl9pc3N1ZXMi\nOjAsIndhdGNoZXJzIjowLCJkZWZhdWx0X2JyYW5jaCI6Im1hc3RlciIsIm1h\nc3Rlcl9icmFuY2giOiJtYXN0ZXIiLCJwZXJtaXNzaW9ucyI6eyJhZG1pbiI6\ndHJ1ZSwicHVzaCI6dHJ1ZSwicHVsbCI6dHJ1ZX19XSwidXJsIjoiaHR0cHM6\nLy9hcGkuZ2l0aHViLmNvbS9vcmdzLzxHSVRIVUJfVEVTVF9PUkdBTklaQVRJ\nT04+L21pZ3JhdGlvbnMvOTciLCJjcmVhdGVkX2F0IjoiMjAxNS0wNy0yOFQx\nODoxMjozMi0wNzowMCIsInVwZGF0ZWRfYXQiOiIyMDE1LTA3LTI4VDE4OjEy\nOjMyLTA3OjAwIn0=\n"},"http_version":null},"recorded_at":"Wed, 29 Jul 2015 01:12:32 GMT"}],"recorded_with":"VCR 2.9.3"}
```
jonmagic: It looks like this is actually missing the repositories parameter in it?

lib/octokit/client/organizations.rb
```diff
@@ -595,6 +595,87 @@ def remove_organization_membership(org, options = {})
...
+      #
+      # @param org [String, Integer] Organization GitHub login or id.
+      # @param repositories [Array<String>] :repositories Repositories for the organization.
+      # @option options [Boolean, optional] :lock_repositories Indicates whether repositories should be locked during migration
+      # @return [Sawyer::Resource] Hash representing the new migration.
+      # @see https://developer.github.com/v3/orgs/teams/#create-team
+      # @example
+      #   @client.start_migration('github', ['github/dotfiles'])
+      # @see https://developer.github.com/v3/orgs/migrations/#start-a-migration
+      def start_migration(org, repositories, options = {})
```
jonmagic: I don't see that the "repositories" parameter is actually used anywhere in this method?

gjtorikian: Ugh, such shame. Looks like this was refactored out of existence: https://github.com/octokit/octokit.rb/commit/14542b81e5108c8d0fc85e8b419dc06dc378bf72

Sending a PR now.

PR #629, opened 2015-08-07
Update documentation to reflect the {state: "all"} option

nilbus:
This applies to both issues and milestones where the default state is "open".
Although undocumented, the "all" option is confirmed to work.

See the API docs:
  * https://developer.github.com/v3/issues/#parameters-1
  * https://developer.github.com/v3/issues/milestones/#parameters

### DIFF
--- a/lib/octokit/client/milestones.rb
+++ b/lib/octokit/client/milestones.rb
@@ -11,7 +11,7 @@ module Octokit
       # @param repository [String, Repository, Hash] A GitHub repository.
       # @param options [Hash] A customizable set of options.
       # @option options [Integer] :milestone Milestone number.
-      # @option options [String] :state (open) State: <tt>open</tt> or <tt>closed</tt>.
+      # @option options [String] :state (open) State: <tt>open</tt>, <tt>closed</tt>, or <tt>all</tt>.
       # @option options [String] :sort (created) Sort: <tt>created</tt>, <tt>updated</tt>, or <tt>comments</tt>.
       # @option options [String] :direction (desc) Direction: <tt>asc</tt> or <tt>desc</tt>.
       # @return [Array<Sawyer::Resource>] A list of milestones for a repository.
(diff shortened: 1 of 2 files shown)


lib/octokit/client/milestones.rb
```diff
@@ -28,7 +28,7 @@ def list_milestones(repository, options = {})
       # @param repository [String, Repository, Hash] A GitHub repository.
       # @param options [Hash] A customizable set of options.
       # @option options [Integer] :milestone Milestone number.
-      # @option options [String] :state (open) State: <tt>open</tt> or <tt>closed</tt>.
+      # @option options [String] :state (open) State: <tt>open</tt>, <tt>closed</tt>, or <tt>all</tt>.
```
pengwynn: Oops, everything but `number` is existing copypasta. Mind removing `:state`, `:sort`, and `:direction here?

lib/octokit/client/milestones.rb
```diff
@@ -44,7 +44,7 @@ def milestone(repository, number, options = {})
       # @param repository [String, Repository, Hash] A GitHub repository.
       # @param title [String] A unique title.
       # @param options [Hash] A customizable set of options.
-      # @option options [String] :state (open) State: <tt>open</tt> or <tt>closed</tt>.
+      # @option options [String] :state (open) State: <tt>open</tt>, <tt>closed</tt>, or <tt>all</tt>.
```
pengwynn: The `all` value works for listings, but isn't a valid value for specifying state on a milestone.

lib/octokit/client/milestones.rb
```diff
@@ -61,7 +61,7 @@ def create_milestone(repository, title, options = {})
       # @param number [String, Integer] ID of the milestone
       # @param options [Hash] A customizable set of options.
       # @option options [String] :title A unique title.
-      # @option options [String] :state (open) State: <tt>open</tt> or <tt>closed</tt>.
+      # @option options [String] :state (open) State: <tt>open</tt>, <tt>closed</tt>, or <tt>all</tt>.
```
pengwynn: Same here, the `all` value works for listings, but isn't a valid value for specifying state on a milestone.

pengwynn:
Thanks @nilbus. I left a couple of comments.

nilbus:
:dizzy_face: good catch! Thank you for looking so closely. I will update this soon.

nilbus:
Updated! This looks better to me now.

pengwynn:
:cake:

PR #632, opened 2015-08-11
DOC list_issues accept `all` as a state

byroot:
See: https://developer.github.com/v3/issues/#list-issues

Regards.


byroot:
My bad I just noticed this is a duplicate of https://github.com/octokit/octokit.rb/pull/629

PR #641, opened 2015-09-17
Add validation of repository format

simon0191:
Hi,

I was using `Octokit.repository?` to check if a string was a valid github repository and I realised that it was returning `true` for repositories like `paths_with/more/than/one/slash`. I thought it could be a good idea to add a validation to the format of the repository and username.

This pull request:
- Validates repository format against `URI.regexp`
- Add `Octokit::InvalidRepository` error

### DIFF
--- a/lib/octokit/repository.rb
+++ b/lib/octokit/repository.rb
@@ -4,6 +4,7 @@ module Octokit
   # URLs and to generate URLs
   class Repository
     attr_accessor :owner, :name, :id
+    NAME_WITH_OWNER_PATTERN = /\A(?<owner>[\.\w-]+)\/(?<name>[\.\w-]+)\z/i
 
     # Instantiate from a GitHub repository URL
     #
@@ -12,15 +13,14 @@ module Octokit
       Repository.new(URI.parse(url).path[1..-1])
     end
 
+    # @raise [Octokit::InvalidRepository] if the repository
+    #   has an invalid format
     def initialize(repo)
       case repo
       when Integer
         @id = repo
-      when String
-        @owner, @name = repo.split('/')
-        unless @owner && @name
-          raise ArgumentError, "Invalid Repository. Use user/repo format."
-        end
+      when NAME_WITH_OWNER_PATTERN
+        @owner, @name = repo.split("/")
       when Repository
         @owner = repo.owner
         @name = repo.name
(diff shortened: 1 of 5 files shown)


lib/octokit/repository.rb
```diff
@@ -71,5 +78,17 @@ def url
     alias :user :owner
     alias :username :owner
     alias :repo :name
+
+    private
+
+      def validate_owner_and_name!
+        if @owner.match(/\//) || @name.match(/\//) || !url.match(/\A#{URI.regexp}\z/)
```
gjtorikian: Let's use [`include`](http://ruby-doc.org/core-1.9.3/String.html#method-i-include-3F) here for the first two checks.

simon0191: :+1:

simon0191:
@gjtorikian thanks for the feedback. I rebased the commit with the suggestion

gjtorikian:
LGTM. @pengwynn ?

lib/octokit/repository.rb
```diff
@@ -12,14 +12,18 @@ def self.from_url(url)
...
         @id = repo
       when String
-        @owner, @name = repo.split('/')
-        unless @owner && @name
-          raise ArgumentError, "Invalid Repository. Use user/repo format."
+        tokens = repo.split('/')
+        if tokens.length == 2
+          @owner, @name = tokens
+        else
+          raise_invalid_repository!
```
pengwynn: I think we could tighten this up even more:

```ruby
NAME_WITH_OWNER_PATTERN = /\A(?<owner>[\.\w-]+)\/(?<name>[\.\w-]+)\z/i

...

case repo
when Integer
  @id = repo
when NAME_WITH_OWNER_PATTERN
  @owner, @name = repo.split("/")
when Repository
...
```

What do you think?

simon0191: :+1:  I like it. 
@pengwynn This is my first pull request for an Open source project. Is ok if I ammend the commit or is it better to make a new one?

pengwynn: @simon0191 either way is fine. Thanks for the patch

simon0191: @pengwynn thanks for the feedback, I amended the commit

pengwynn:
:cake:

PR #644, opened 2015-09-23
Merge authorization header with options

### LINKED ISSUE #643: Pass additional headers into Octokit::Client::Authorizations#scope
Currently the [scopes method](https://github.com/octokit/octokit.rb/blob/master/lib/octokit/client/authorizations.rb#L118) doesn't allow us to pass additional headers for the request.

My use case is that I want to set the

```ruby
'Cache-Control' => 'no-cache, no-store'
```

header so that the response isn't cached.

Is this a change that would be welcome?

/cc @johndbritton

tarebyte:
Closes #643, this allows the user to pass additional options to the `scopes` method.

Feedback welcome

/cc @johndbritton @pengwynn

### DIFF
--- a/lib/octokit/client/authorizations.rb
+++ b/lib/octokit/client/authorizations.rb
@@ -113,6 +113,7 @@ module Octokit
       # Check scopes for a token
       #
       # @param token [String] GitHub OAuth token
+      # @param options [Hash] A customizable set of options.
       # @return [Array<String>] OAuth scopes
       # @see https://developer.github.com/v3/oauth/#scopes
       def scopes(token = @access_token, options = {})
--- a/lib/octokit/client/authorizations.rb
+++ b/lib/octokit/client/authorizations.rb
@@ -113,20 +113,16 @@ module Octokit
       # Check scopes for a token
       #
       # @param token [String] GitHub OAuth token
-      # @param options [Hash] A customizable set of options.
+      # @param options [Hash] Header params for request
       # @return [Array<String>] OAuth scopes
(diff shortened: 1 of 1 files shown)


lib/octokit/client/authorizations.rb
```diff
@@ -115,10 +115,17 @@ def delete_authorization(number, options = {})
       # @param token [String] GitHub OAuth token
```
johndbritton: > @param options [Hash] ...

lib/octokit/client/authorizations.rb
```diff
@@ -113,12 +113,20 @@ def delete_authorization(number, options = {})
...
-      def scopes(token = @access_token)
+      def scopes(token = @access_token, options = {})
         raise ArgumentError.new("Access token required") if token.nil?
+        authorization_header = { "Authorization" => "token #{token}" }
+
+        if options[:headers]
+          options[:headers].merge(authorization_header)
+        else
+          options[:headers] = authorization_header
+        end
```
pengwynn: Just a style preference, I think I'd tackle this like:

```ruby
auth = { "Authorization" => "token #{token}" }
headers = (options.delete(:headers) || {}).merge(auth)

agent.call(:get, "user", :headers => headers).

tarebyte:
I think this is ready to go unless anyone else has anymore feedback :smile:

pengwynn:
:cake:

tarebyte:
Thanks @pengwynn :heart_eyes:

PR #652, opened 2015-10-16
Add license_contents method

benbalter:
This adds a `license_contents` method to hit the [license contents endpoint](https://developer.github.com/v3/licenses/#get-a-repositorys-license) and return the content of a repository's license file.

//cc @bkeepers


bkeepers:
:+1: LGTM

lib/octokit/client/licenses.rb
```diff
@@ -27,6 +27,19 @@ def license(license_name, options = {})
...
+      #
+      # @see https://developer.github.com/v3/licenses/#get-a-repositorys-license
+      # @param repo [Integer, String, Repository, Hash] A GitHub repository
+      # @option options [String] :ref name of the Commit/Branch/Tag. Defaults to “master”.
+      # @return [Sawyer::Resource] The detail of the license file
+      # @example
+      #   Octokit.license_contents 'benbalter/licensee'
+      def license_contents(repo, options = {})
+        options = ensure_api_media_type(:licenses, options)
+        get "#{Repository.path repo}/license", options
```
pengwynn: We should probably call this `repository_license_contents`. If the content is Base64, can we decode that before returning to the caller?

benbalter: > We should probably call this repository_license_contents

Done via 99427a9.

> If the content is Base64, can we decode that before returning to the caller?

We can, the other content methods (e.g., README, doesn't as far as I can see).

Related, should this be in the license file or with the other content getters?

bkeepers: >>  If the content is Base64, can we decode that before returning to the caller?

> We can, the other content methods (e.g., README, doesn't as far as I can see).

I was actually thrown off by the other methods not doing this the other day, especially since you have to call `.force_encoding("UTF-8")` after decoding it.

pengwynn:
Moving [this conversation](https://github.com/octokit/octokit.rb/pull/652#discussion_r42255532) out of the outdated diff:

> Related, should this be in the license file or with the other content getters?

@benbalter I think here is fine.

> We can, the other content methods (e.g., README, doesn't as far as I can see).
> I was actually thrown off by the other methods not doing this the other day, especially since you have to call .force_encoding("UTF-8") after decoding it.

@benbalter @bkeepers: Thanks for pointing that out. I guess it would break compatibility to change that behavior now. Maybe go with what we've got now and change that in Octokit v5?

bkeepers:
> Thanks for pointing that out. I guess it would break compatibility to change that behavior now. Maybe go with what we've got now and change that in Octokit v5?

:+1: 

Another option would be to add a `Octokit.decode_content = false` option that gets flipped to `true` in v5?

pengwynn:
> Another option would be to add a Octokit.decode_content = false option that gets flipped to true in v5?

@bkeepers I dig that. 

@benbalter: mind tackling that in this or another PR?

lib/octokit/client/licenses.rb
```diff
@@ -27,6 +27,19 @@ def license(license_name, options = {})
         options = ensure_api_media_type(:licenses, options)
         get "licenses/#{license_name}", options
       end
+
+      # Returns the contents of the repository’s license file, if one is detected.
+      #
+      # @see https://developer.github.com/v3/licenses/#get-a-repositorys-license
+      # @param repo [Integer, String, Repository, Hash] A GitHub repository
+      # @option options [String] :ref name of the Commit/Branch/Tag. Defaults to “master”.
```
JuanitoFatas: `Defaults to “master”.` The quotes should be `"`?

pengwynn: Great catch @JuanitoFatas.

benbalter: Fixed via 37d2479.

PR #658, opened 2015-11-10
Add specific error when user has unverified email (for 403s)

johnpaulashenfelter:
A user with an unverified email address gets a 403 error when trying
to create a repository through the API. This can easily happen with a
brand new user account.

Some of the 403 errors (abuse, rate limits, etc) already have their
own error subclass to make it easier to work with -- this PR simply
adds a specific Octokit::UnverifiedEmail instead of relying on the
more generic Octokit::Forbidden error.

### DIFF
--- a/lib/octokit/error.rb
+++ b/lib/octokit/error.rb
@@ -66,6 +66,8 @@ module Octokit
         Octokit::AbuseDetected
       elsif body =~ /repository access blocked/i
         Octokit::RepositoryUnavailable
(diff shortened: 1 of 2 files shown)


matthewmccullough:
This would be helpful for a Training platform/app/framework where having verified Email addresses are an important part of the flow in interaction with the students.

pengwynn:
Thanks! :cake:

PR #661, opened 2015-11-12
Ensure the correct preview warning message is issued

pvdb:
@pengwynn, @spraints, @jonmagic, _et al._ - this PR fixes the problem I documented in #627, namely that the message issued by the `Octokit::Preview.warn_preview` method is always warning about the "Migrations API", irrespective of the preview type passed into the `Octokit::Preview.ensure_api_media_type` method... small change, big difference! :smile:  thanks!

---

**before:**

```ruby
[1] pry(main)> Octokit::Preview.send(:module_function, :warn_preview)
=> Octokit::Preview
[2] pry(main)> Octokit::Preview.send(:module_function, :ensure_api_media_type)
=> Octokit::Preview
[3] pry(main)> Octokit::Preview.ensure_api_media_type(:migrations, {})
WARNING: The preview version of the Migrations API is not yet suitable for production use.
You can avoid this message by supplying an appropriate media type in the 'Accept' request
header.
=> {:accept=>"application/vnd.github.wyandotte-preview+json"}
[4] pry(main)> Octokit::Preview.ensure_api_media_type(:licenses, {})
WARNING: The preview version of the Migrations API is not yet suitable for production use.
You can avoid this message by supplying an appropriate media type in the 'Accept' request
header.
=> {:accept=>"application/vnd.github.drax-preview+json"}
[5] pry(main)> _
```

**after:**

```ruby
[1] pry(main)> Octokit::Preview.send(:module_function, :warn_preview)
=> Octokit::Preview
[2] pry(main)> Octokit::Preview.send(:module_function, :ensure_api_media_type)
=> Octokit::Preview
[3] pry(main)> Octokit::Preview.ensure_api_media_type(:migrations, {})
WARNING: The preview version of the Migrations API is not yet suitable for production use.
You can avoid this message by supplying an appropriate media type in the 'Accept' request
header.
=> {:accept=>"application/vnd.github.wyandotte-preview+json"}
[4] pry(main)> Octokit::Preview.ensure_api_media_type(:licenses, {})
WARNING: The preview version of the Licenses API is not yet suitable for production use.
You can avoid this message by supplying an appropriate media type in the 'Accept' request
header.
=> {:accept=>"application/vnd.github.drax-preview+json"}
[5] pry(main)> _
```

PR #662, opened 2015-11-12
Support the "branch protection" preview API (`loki`)

pvdb:
@pengwynn, @spraints, @jonmagic, _et al._ - this PR adds support for the "branch protection" preview API - code-named `loki-preview` - which was announced yesterday:

https://developer.github.com/changes/2015-11-11-protected-branches-api/

Use of `loki-preview` is left optional and up to the user, but - if used - retrieves the `protection` data of a branch, as illustrated by this example:

```shell
[1] pry(main)> ENV['OCTOKIT_ACCESS_TOKEN'].empty?
=> false
[2] pry(main)> pry Octokit::Client.new
[1] pry(#<Octokit::Client>)> ensure_api_media_type(:branch_protection, {})
WARNING: The preview version of the Migrations API is not yet suitable for production use.
You can avoid this message by supplying an appropriate media type in the 'Accept' request
header.
=> {:accept=>"application/vnd.github.loki-preview+json"}
[2] pry(#<Octokit::Client>)> # WITHOUT the "branch protection" preview
[3] pry(#<Octokit::Client>)> get_branch('octokit/octokit.rb', 'master')[:protection]
=> nil
[4] pry(#<Octokit::Client>)> # WITH the "branch protection" preview
[5] pry(#<Octokit::Client>)> get_branch('octokit/octokit.rb', 'master', ensure_api_media_type(:branch_protection, {}))[:protection]
WARNING: The preview version of the Migrations API is not yet suitable for production use.
You can avoid this message by supplying an appropriate media type in the 'Accept' request
header.
=> {:enabled=>false, :required_status_checks=>{:enforcement_level=>"off", :contexts=>[]}}
[6] pry(#<Octokit::Client>)> _
```

---

Note that the warning messages in the above example are wrong due the a bug introduced in #627, but will be correct once #661 gets merged in.

### DIFF
--- a/lib/octokit/preview.rb
+++ b/lib/octokit/preview.rb
@@ -4,8 +4,9 @@ module Octokit
   module Preview
 
     PREVIEW_TYPES = {
-      :migrations    => 'application/vnd.github.wyandotte-preview+json'.freeze,
-      :licenses      => 'application/vnd.github.drax-preview+json'.freeze
+      :branch_protection => 'application/vnd.github.loki-preview+json'.freeze,
+      :migrations        => 'application/vnd.github.wyandotte-preview+json'.freeze,
+      :licenses          => 'application/vnd.github.drax-preview+json'.freeze
     }
 
     def ensure_api_media_type(type, options)


gjtorikian:
Hey @pvdb—thanks for this PR! Would you mind adding a test and checking in [the corresponding cassette](https://github.com/octokit/octokit.rb/tree/master/spec/cassettes) ?

pvdb:
Thanks for reviewing the PR, @gjtorikian, glad you like it... it's gonna be very useful for us, for querying our org repos in a scripted and automated fashion, as we're quite keen on leveraging the branch protection features.

You wrote:
> Would you mind adding a test and checking in the corresponding cassette ?

I wouldn't, but when I issued the PR, I looked through the existing specs but couldn't find any that explicitly test the functionality provided by the `Octokit::Preview` module, most notably the `Octokit::Preview#ensure_api_media_type` method... it is only ever tested indirectly via the `Octokit::Client::Licenses` module, AFAICT, but never directly

Having said that, though, I'll have a look at the specs you wrote for #627 and have a stab at specs for the `loki` preview API, maybe testing it indirectly via the branch specs in `repositories_spec.rb`... thoughts?

pengwynn:
@pvdb This patch looks great, thanks. I can go ahead and merge it, but to really implement everything needed for the title of the PR, I think we need new methods for `protect_branch` and `unprotect_branch` that would call [the new `PATCH` method][patch]. These methods would need the `ensure_api_media_type` since they're preview-only.

Would you like to do that here or should we open a new PR?

[patch]: https://developer.github.com/v3/repos/#enabling-and-disabling-branch-protection

pvdb:
Thanks for accepting the PR, @pengwynn, and thanks @tarebyte for implementing the two missing methods! #so_much_win :smile:

PR #664, opened 2015-11-17
Implement branch protection methods

tarebyte:
From @pengwynn's comment in #662 I implemented the `protect_branch` and `unprotect_branch` methods for the [new protected-branches api](https://developer.github.com/changes/2015-11-11-protected-branches-api/)

Feedback would be very much welcome :smile_cat:


pengwynn:
Fantastic, looks great.  Thanks @tarebyte. :clap:

tarebyte:
:sparkles:

pvdb:
So much win, @tarebyte ... thanks!  I was about to have a go at implementing those two methods, but noticed just in time that you already had!  :thumbsup:

PR #667, opened 2015-11-17
Fix org creation to include login and admin.

nmaki:
Login and Admin need to be added to the post options to properly create an org.


gjtorikian:
Thank you! :pineapple:

PR #668, opened 2015-11-19
Allow org hooks to be used by org login or ID

tarebyte:
This uses `Organization.path` so that we can access the org webooks by `login` or `ID`

/cc @pengwynn @kdaigle


kdaigle:
:sparkles: Thanks @tarebyte!

pengwynn:
:cake:

PR #673, opened 2015-11-30
Implement add or update organization membership

raysrashmi:
Implement add or update organization membership
https://developer.github.com/v3/orgs/members/#add-or-update-organization
-membership


tarebyte:
@raysrashmi `#update_organization_membership` already does this: https://github.com/octokit/octokit.rb/blob/a2b1217f9662b7ef2ab718ec32f54283c9562546/lib/octokit/client/organizations.rb#L579

But I can see why the method name could be misleading

raysrashmi:
@tarebyte Oh I missed that. Thanks for the confirmation

PR #677, opened 2015-12-05
Actually pass `:repositories` over to the migration API

gjtorikian:
Noted by @jonmagic over at https://github.com/octokit/octokit.rb/pull/627#discussion-diff-46722760.

This doesn't change the method signature so it shouldn't be considered a breaking change. Sorry for the trouble. :cold_sweat:


pengwynn:
:+1:

jonmagic:
@gjtorikian awesome! Thanks for the quick fix :metal:

Issue #671, opened 2015-11-25
warning: method redefined; discarding old management_console_endpoint


pvdb:
There's something really funky going on with the `management_console_endpoint` method, which is defined in two places:

* `Octokit::Default.management_console_endpoint`
* `Octokit::Configurable#management_console_endpoint`

To illustrate the issue:

```shell
$ ruby -W2 -I lib -r octokit -e exit
~/octokit.rb/lib/octokit/configurable.rb:111: warning: method redefined; discarding old management_console_endpoint
$ _
```

I can't say I understand the _(TBH quite convoluted)_ code, but once I do I'll issue a PR, as one of the two versions is most likely superfluous and/or unused.

pengwynn:
@pvdb one of those modules provides the configurable Octokit settings (available both as [module-level and instance-level settings][settings]), the other provides the default values for those settings. In this case, we may be able to call `reset_agent` in `Octokit::Configurable#management_console_endpoint` and drop `Octokit::EnterpriseManagementConsoleClient#management_console_endpoint`.

Thoughts @gjtorikian?

[settings]: https://github.com/octokit/octokit.rb#configuration-and-defaults

gjtorikian:
I believe we can go even simpler. Here's a diff that shuffles the attribute to from `accesor` to simply `writer`:

``` diff
diff --git a/lib/octokit/configurable.rb b/lib/octokit/configurable.rb
index a377523..b97a6c0 100644
--- a/lib/octokit/configurable.rb
+++ b/lib/octokit/configurable.rb
@@ -49,10 +49,10 @@ module Octokit
 
     attr_accessor :access_token, :auto_paginate, :client_id,
                   :client_secret, :default_media_type, :connection_options,
-                  :management_console_endpoint, :management_console_password,
+                  :management_console_password,
                   :middleware, :netrc, :netrc_file,
                   :per_page, :proxy, :user_agent
-    attr_writer :password, :web_endpoint, :api_endpoint, :login
+    attr_writer :password, :web_endpoint, :api_endpoint, :management_console_endpoint, :login
 
     class << self
```

This, including the separate `Configurable` and `Client` methods, is exactly the same way `api_endpoint` and `web_endpoint` are defined. I suspect we were trying to model those definitions but missed this attribute distinction. 

With this diff the warning goes away. It looks like `management_console_password` can also drop down to `writer`, to match it counterpart, `:password`.

pengwynn:
Awesome, that should work. :+1:

PR #678, opened 2015-12-07
Set `management_console_*` ivars as only writeable

gjtorikian:
This should close https://github.com/octokit/octokit.rb/issues/671.

/cc @pvdb It'd be great if you could grab this branch and confirm there's no warning, if possible.
/c @pengwynn


pvdb:
Hey @gjtorikian,

You wrote:
> /cc @pvdb It'd be great if you could grab this branch and confirm there's no warning, if possible.

I can confirm that - using my little snippet from #671 - the re-definition warning message is NOT issued against the code on this branch :thumbs: :shipit:

PR #679, opened 2015-12-12
Fix documentation urls

joeyw:
Fix a number of documentation urls


pengwynn:
Sweet. Did you check these manually or do you have a script?

joeyw:
Half and half. I used the chrome extension I made a while back https://github.com/joeyw/octodoctotron and just browsed the api to see which ones were missing/misplaced.

Issue #685, opened 2015-12-18
Add stats timeout configuration options


kv109:
Hi!

Github API [statistics](https://developer.github.com/v3/repos/statistics/) call returns `202` if they are not calculated yet. I think it would be useful to add `stats_timeout` configuration option, which would tell Octokit to try to retrieve this data for a given amount of time. PR could look like [this](
https://github.com/octokit/octokit.rb/compare/master...kv109:master?diff=unified&name=master).

What do you think?

PS. Thank you for this project, it works amazing.

pengwynn:
@kv109 the idea is intriguing. Wanna open a PR so we could have a conversation on the diff?

kv109:
@pengwynn With pleasure: https://github.com/octokit/octokit.rb/pull/687. No idea how to test this functionality, though.

PR #687, opened 2015-12-19
Fetch stats via retry and timeout

kv109:
Github API [statistics](https://developer.github.com/v3/repos/statistics/) call returns `202` if they are not calculated yet. This PR adds the ability to automatically retry fetching those stats.

### DIFF
--- a/lib/octokit/client/stats.rb
+++ b/lib/octokit/client/stats.rb
@@ -75,11 +75,32 @@ module Octokit
       #
       # @param repo [Integer, String, Hash, Repository] A GitHub repository
       # @param metric [String] The metrics you are looking for
-      # @return [Array<Sawyer::Resource>] Magical unicorn stats
+      # @return [Array<Sawyer::Resource> or nil] Stats in metric-specific format, or nil if not yet calculated.
+      # @see https://developer.github.com/v3/repos/statistics/
       def get_stats(repo, metric, options = {})
-        data = get("#{Repository.path repo}/stats/#{metric}", options)
+        path = "#{Repository.path repo}/stats/#{metric}"
+        if retry_timeout = options.delete(:retry_timeout)
+          get_stats_with_retry(retry_timeout, path, options)
+        else
+          get(path, options)
+        end.tap do
+          return nil if last_response.status == 202
+        end
+      end
 
-        last_response.status == 202 ? nil : data
+      # @private Get stats for a repository. Retry call until stats are ready or retry_timeout exceeded.
+      #
+      # @param retry_timeout [String] For how long Octokit should keep trying to get stats (in seconds)
+      # @return [Array<Sawyer::Resource> or nil] Stats in metric-specific format, or nil if not yet calculated.
+      def get_stats_with_retry(retry_timeout, path, options = {})
+        time_start = Time.now
+
+        loop do
+          data = get(path, options)
+          timeout_exceeded = Time.now - time_start >= retry_timeout
+          return data if last_response.status != 202 || timeout_exceeded
+          sleep retry_timeout / 4.0
+        end
       end
     end
   end


lib/octokit/client/stats.rb
```diff
@@ -75,11 +75,26 @@ def punch_card_stats(repo, options = {})
       #
       # @param repo [Integer, String, Hash, Repository] A GitHub repository
       # @param metric [String] The metrics you are looking for
-      # @return [Array<Sawyer::Resource>] Magical unicorn stats
+      # @return [Array<Sawyer::Resource> or nil] Magical unicorn stats or nil if unicorn haven't calculated stats yet
```
pengwynn: Ugh, the original return type comment here is bad. We should make this:

> Stats in metric-specific format, or nil if not yet calculated.

... and add a `@see https://developer.github.com/v3/repos/statistics/`

lib/octokit/client/stats.rb
```diff
@@ -75,11 +75,26 @@ def punch_card_stats(repo, options = {})
...
+          return nil if last_response.status == 202
+        end
+      end
+
+      # @private Get stats for a repository. Repeat call to Github until stats are ready or [Configurable#stats_timeout] exceeded.
+      #
+      # @param repo [Integer, String, Hash, Repository] A GitHub repository
+      # @param metric [String] The metrics you are looking for
+      # @return [Array<Sawyer::Resource>] Magical unicorn stats
+      def get_stats_data_patiently(repo, metric, options = {})
```
pengwynn: Maybe `get_stats_with_retry`?

lib/octokit/client/stats.rb
```diff
@@ -75,11 +75,26 @@ def punch_card_stats(repo, options = {})
       #
       # @param repo [Integer, String, Hash, Repository] A GitHub repository
       # @param metric [String] The metrics you are looking for
-      # @return [Array<Sawyer::Resource>] Magical unicorn stats
+      # @return [Array<Sawyer::Resource> or nil] Magical unicorn stats or nil if unicorn haven't calculated stats yet
       def get_stats(repo, metric, options = {})
-        data = get("#{Repository.path repo}/stats/#{metric}", options)
+        get_stats_data_patiently(repo, metric, options).tap do
+          return nil if last_response.status == 202
```
pengwynn: Instead of always retrying, I'd rather opt-in. I'd also like to avoid yet another configuration option. 

What if this looked like (with hypothetical `Client#get_with_retry` method):

```ruby
def get_stats(repo, metric, options = {})
  path = "#{Repository.path repo}/stats/#{metric}"
  if retry_timeout = options.delete(:retry_timeout)
    get_with_retry(path, options)
  else
    get(path, options)
  end
end
```
We could make the logic in `get_with_retry` explicitly look for a `202` since I _think_ that's the only case where we'd want to retry blindly.

pengwynn:
@kv109 thanks for the pull. I left a couple of notes. You can push your changes to the same branch and they'll update here.

pengwynn:
@kv109 I also added a PR description, riffing on what you wrote in #685. :bow:

lib/octokit/client/stats.rb
```diff
@@ -75,11 +75,26 @@ def punch_card_stats(repo, options = {})
...
+      # @param metric [String] The metrics you are looking for
+      # @return [Array<Sawyer::Resource>] Magical unicorn stats
+      def get_stats_data_patiently(repo, metric, options = {})
+        time_start = Time.now
 
-        last_response.status == 202 ? nil : data
+        loop do
+          data = get("#{Repository.path repo}/stats/#{metric}", options)
+          return data if last_response.status == 200 || Time.now - time_start >= stats_timeout
+          sleep 0.1
```
joeyw: Delay between retries could be an issue here. Even if the API takes a few seconds to generate the stats, at `0.1` this could add up to a lot of wasted requests. It should probably be longer. Not to make it too complicated, but maybe it could delay longer and longer in a number of steps until it reaches the set timeout.

kv109:
@pengwynn @joeyw 
Code review applied (hopefully)

lib/octokit/client/stats.rb
```diff
@@ -75,11 +75,32 @@ def punch_card_stats(repo, options = {})
...
+      # @return [Array<Sawyer::Resource> or nil] Stats in metric-specific format, or nil if not yet calculated.
+      def get_stats_with_retry(retry_timeout, path, options = {})
+        time_start = Time.now
 
-        last_response.status == 202 ? nil : data
+        loop do
+          data = get(path, options)
+          timeout_exceeded = Time.now - time_start >= retry_timeout
+          return data if last_response.status != 202 || timeout_exceeded
+          sleep retry_timeout / 3.0
```
pengwynn: Curious, why divide by 3 here?

lib/octokit/client/stats.rb
```diff
@@ -75,11 +75,32 @@ def punch_card_stats(repo, options = {})
...
+        end.tap do
+          return nil if last_response.status == 202
+        end
+      end
+
+      # @private Get stats for a repository. Retry call until stats are ready or retry_timeout exceeded.
+      #
+      # @param retry_timeout [String] For how long Octokit should keep trying to get stats (in milliseconds)
+      # @return [Array<Sawyer::Resource> or nil] Stats in metric-specific format, or nil if not yet calculated.
+      def get_stats_with_retry(retry_timeout, path, options = {})
```
pengwynn: I'm fine with leaving this here until we prove the concept, but this is generic enough we could perhaps move to `Client` as `Client#get_with_retry`.

lib/octokit/client/stats.rb
```diff
@@ -75,11 +75,32 @@ def punch_card_stats(repo, options = {})
...
+      # @see https://developer.github.com/v3/repos/statistics/
       def get_stats(repo, metric, options = {})
-        data = get("#{Repository.path repo}/stats/#{metric}", options)
+        path = "#{Repository.path repo}/stats/#{metric}"
+        if retry_timeout = options.delete(:retry_timeout)
+          get_stats_with_retry(retry_timeout, path, options)
+        else
+          get(path, options)
+        end.tap do
+          return nil if last_response.status == 202
```
pengwynn: Do we still need the `tap` here?

kv109: With retry or not, we can still end up with `202`. `#get` returns an empty string in such case and we need to convert this empty string to `nil`. At least this is what old version did, see [this line](https://github.com/octokit/octokit.rb/pull/687#discussion-diff-48182966L82).

lib/octokit/client/stats.rb
kv109: To make sure that there will be no more than two retries. But now I think that three retries will be more suitable, so I've changed denominator to `4`. For better understanding:
```
> Octokit.contributors_stats('octokit/octokit.rb', retry_timeout: 10)
!starting #get_stats_with_retry with retry_timeout: 10 sec!

sending call... (1)
time left: 6.8 sec
- sleeping for 2.5 sec...
sending call... (2)
time left: 0.6 sec
- sleeping for 2.5 sec...
sending call... (3)
time left: -6.7 sec: TIMEOUT EXCEEDED

!stats still not calculated, returning nil!
```

lib/octokit/client/stats.rb
kv109: I'd be proud & happy, but I'm a little afraid that I'm not able to write a proper documentation for such generic method.

kv109:
Questions answered

kv109:
@pengwynn How about making a Christmas gift for me and finish this? ;)

Merry Christmas!

giphy

PR #688, opened 2015-12-20
Support getting Gist at specific revision

### LINKED ISSUE #681: Gist at specific revision
https://developer.github.com/v3/gists/#get-a-specific-revision-of-a-gist is currently unsupported in octokit. Should [#gist](https://github.com/octokit/octokit.rb/blob/28a594ff371ff3faf8ec50dd25bd65bd779ff43f/lib/octokit/client/gists.rb#L50) be extended to take the revision SHA or add a new method specifically for it?

joeyw:
Change `.gist` to take `:sha` to get a gist at a specific revision Fix #681


lib/octokit/client/gists.rb
```diff
@@ -45,10 +45,16 @@ def starred_gists(options = {})
       # Get a single gist
       #
       # @param gist [String] ID of gist to fetch
+      # @option options [String] :sha Specific gist revision SHA
       # @return [Sawyer::Resource] Gist information
       # @see https://developer.github.com/v3/gists/#get-a-single-gist
+      # @see https://developer.github.com/v3/gists/#get-a-specific-revision-of-a-gist
       def gist(gist, options = {})
-        get "gists/#{Gist.new(gist)}", options
+        if sha = options.delete(:sha)
```
joeyw: Would `:sha` be better named `:revision` or `:revision_sha`?

gjtorikian: Good question. Since [the docs](https://developer.github.com/v3/gists/#get-a-specific-revision-of-a-gist) also use `:sha` I guess it's okay to use that here too.

pengwynn: > Since the docs also use :sha I guess it's okay to use that here too.

Yeah I like that.

pengwynn:
:cake:

PR #689, opened 2015-12-20
Remove autoprepended data from .create_ref error messages 

### LINKED ISSUE #675: create_ref error message can be unhelpful
`Octokit::Client.create_ref` will return a message from the API about how to format refs verbatim:

```
POST https://api.github.com/repos/foo/bar/git/refs: 422 - Reference name must contain at least three slash-separated components. // See: https://developer.github.com/v3/git/refs/#create-a-reference
```

But this message isn't very helpful, because Octokit will prepend `refs/` to anything given. If one follows the error message's advice, sending in `refs/heads/whatever`, the result is:

```
POST https://api.github.com/repos/foo/bar/git/refs: 422 - refs//refs/heads/potential-master-foo is not a valid ref name. // See: https://developer.github.com/v3/git/refs/#create-a-reference
```

Since we're going to pass the error message back to the client, and we're going to edit the data the client gives us, should we strip a leading `refs/` from client input? I'm don't think `refs/refs` is ever valid anyway, and it would make following the error message's advice work.

joeyw:
Remove "refs/" that is added by octokit automatically from any error messages

Fixes #675

### DIFF
--- a/lib/octokit/client/refs.rb
+++ b/lib/octokit/client/refs.rb
@@ -46,8 +46,9 @@ module Octokit
       # @example Create refs/heads/master for octocat/Hello-World with sha 827efc6d56897b048c772eb4087f854f46256132
       #   Octokit.create_ref("octocat/Hello-World","heads/master", "827efc6d56897b048c772eb4087f854f46256132")
       def create_ref(repo, ref, sha, options = {})
+        ref = "refs/#{ref}" unless ref =~ %r{refs/}
         parameters = {
-          :ref  => "refs/#{ref}",
+          :ref  => ref,
           :sha  => sha
         }
         post "#{Repository.path repo}/git/refs", options.merge(parameters)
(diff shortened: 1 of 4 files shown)


gjtorikian:
Hmm. Is `refs/` an addition from Octokit, or the actual API? If the latter it seems the problem should be fixed upstream.

joeyw:
It's added by octokit https://github.com/octokit/octokit.rb/blob/28a594ff371ff3faf8ec50dd25bd65bd779ff43f/lib/octokit/client/refs.rb#L50

pengwynn:
Thanks, @joeyw. I think we were wanting to fix this on the input side. Something like:

```ruby
def create_ref(repo, ref, sha, options = {})
  ref = "refs/#{ref}" unless ref =~ %r(refs/)
  parameters = {
    :ref  => ref,
    :sha  => sha
  }
  ...
```

joeyw:
Well that is a lot easier. Nothing like fixing the wrong problem :)

pengwynn:
Looks good @joeyw. :cake:

bhuga:
:heart:

PR #692, opened 2015-12-25
Fetch stats with retry

### LINKED ISSUE #685: Add stats timeout configuration options

pengwynn:
Riffing on #687, this takes a test-driven approach to adding a retry option to fetching stats.

Merry Christmas @kv109 :gift: :christmas_tree: https://github.com/octokit/octokit.rb/pull/687#issuecomment-167116234

/cc @joeyw for :eyes: 

Closes #685, closes #687

### DIFF
--- a/lib/octokit/client/stats.rb
+++ b/lib/octokit/client/stats.rb
@@ -9,6 +9,8 @@ module Octokit
       # Get contributors list with additions, deletions, and commit counts
       #
       # @param repo [Integer, String, Hash, Repository] A GitHub repository
+      # @option retry_timeout [Number] How long Octokit should keep trying to get stats (in seconds)
+      # @option retry_wait [Number] How long Octokit should wait between retries.
       # @return [Array<Sawyer::Resource>] Array of contributor stats
       # @see https://developer.github.com/v3/repos/statistics/#get-contributors-list-with-additions-deletions-and-commit-counts
       # @example Get contributor stats for octokit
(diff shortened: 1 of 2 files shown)


lib/octokit/client/stats.rb
```diff
@@ -75,11 +85,36 @@ def punch_card_stats(repo, options = {})
...
+      # @return [Array<Sawyer::Resource> or nil] Stats in metric-specific format, or nil if not yet calculated.
+      def get_stats_with_retry(path, retry_timeout, retry_wait, options = {})
+        time_start = Time.now
 
-        last_response.status == 202 ? nil : data
+        loop do
+          data = get(path, options)
+          timeout_exceeded = Time.now - time_start >= retry_timeout
+          return data if last_response.status != 202 || timeout_exceeded
+          sleep retry_wait
```
joeyw: The timeout doesn't need to be recalcuated each iteration, I think we can simply this further with the following changes:

```ruby
      def get_stats_with_retry(path, retry_timeout, retry_wait, options = {})
        time_out = Time.now + retry_timeout

        loop do
          data = get(path, options)
          return data if last_response.status != 202
          return data if Time.now >= time_out
          sleep retry_wait
        end
      end
```

joeyw: Also, what if we got rid of the 2 helper functions and used 1 instead to get the stats? Since this is an internal api we only care about behaviour, we could add the retry options as parameters instead Something like:

```ruby
def punch_card_stats(repo, options = {})
  get_stats(repo, "punch_card", options, options.delete(:retry_timeout), options.delete(:retry_wait))
end

def get_stats(repo, metric, options = {}, retry_timeout=nil, retry_wait=0.5)
  path = "#{Repository.path repo}/stats/#{metric}"
  timeout = Time.now + retry_timeout if retry_timeout
  loop do
    data = get(path, options)
    return data if last_response.status != 202
    return nil unless retry_timeout
    return data if Time.now >= timeout
    sleep retry_wait if retry_wait
  end
end
```

joeyw:
:metal:

Issue #704, opened 2016-01-05
Would like a way to get an Octokit::Repository from an issue or PR Hash


aroben:
I'm writing a script to show me all PRs opened by me in the last year, and I want to group the PRs by repository. I'm using `Octokit::Client#search_issues` to get the list of PRs. To group them by repository, I need to figure out what repository each PR is in.

The Hash returned by https://developer.github.com/v3/search/#search-issues doesn't include any direct information about the repository. `html_url` of course includes the repository's name-with-owner, so I can use a regex or similar to extract that. It would be nicer if something like `Octokit::Repository.from_url` could do that for me (right now that API only accepts repository URLs, not issue/PR URLs).

Another option would be for the API response to include some information about the repository. That change would happen on github.com, not in Octokit, of course.

joeyw:
Added in #704

PR #705, opened 2016-01-06
Support API urls in Repository.from_url method #704

joeyw:
Parse repositories from api urls in Repository.from_url method /cc @pengwynn @aroben #704

### DIFF
--- a/lib/octokit/repository.rb
+++ b/lib/octokit/repository.rb
@@ -10,11 +10,9 @@ module Octokit
     #
     # @return [Repository]
     def self.from_url(url)
-      repo = URI.parse(url).path[1..-1]
-      if /repos\/(?<owner>[\w.-]+)\/(?<name>[\w.-]+)/i =~ repo
-        repo = {owner: owner, name: name}
-      end
-      new repo
+      new URI.parse(url).path[1..-1].
+        gsub('repos/','').
+        split('/')[0..1]
     end
 
     # @raise [Octokit::InvalidRepository] if the repository
--- a/lib/octokit/repository.rb
+++ b/lib/octokit/repository.rb
@@ -10,7 +10,11 @@ module Octokit
     #
     # @return [Repository]
     def self.from_url(url)
-      Repository.new(URI.parse(url).path[1..-1])
+      repo = URI.parse(url).path[1..-1]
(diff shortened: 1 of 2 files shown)


lib/octokit/repository.rb
```diff
@@ -10,7 +10,11 @@ class Repository
     #
     # @return [Repository]
     def self.from_url(url)
-      Repository.new(URI.parse(url).path[1..-1])
+      repo = URI.parse(url).path[1..-1]
+      if /repos\/(?<owner>[\w.-]+)\/(?<name>[\w.-]+)/i =~ repo
```
pengwynn: Can we move this to the top of the class as a constant? `REPOSITORY_API_URL_PATTERN` maybe?

From looking at #704 I think @aroben was hoping to match on a web site URL, too (`/octokit/octokit.rb/pull/1` vs `/repos/octokit/octokit.rb/pulls/1`). That gets tricky because we can't trust every URL to be a repository, but maybe we don't care.

joeyw:
Round two, now with less regex.

pengwynn:
Looks good @joeyw.

@aroben before I merge this can you see if this would meet your need?

lib/octokit/repository.rb
```diff
@@ -10,7 +10,10 @@ class Repository
     #
     # @return [Repository]
     def self.from_url(url)
-      Repository.new(URI.parse(url).path[1..-1])
+      new URI.parse(url).path[1..-1].
+        gsub('repos/','').
```
aroben: This seems unsafe if, e.g., I have a repository named `aroben/all-my-repos/`. I wonder if it would be safer to have a few defined formats we're willing to accept (perhaps defined by regexes) and explicitly matching against each format?

lib/octokit/repository.rb
```diff
@@ -10,7 +10,10 @@ class Repository
...
+        split('/')[0..1].
```
aroben: You can save a little work by making this `split('/', 3)[0..1].`. That way there won't be any more splitting after the second `/`.

lib/octokit/repository.rb
joeyw: I can change this to only match the start of the path.

PR #710, opened 2016-01-14
Update docs for #list_issues :labels parameter

mastahyeti:
The API accepts a comma separated list of labels, not an Array of Strings. There's a few other methods in this file that claim to accept `[Array<String>]` for the labels parameter. I haven't checked if those need to be updated as well.


pengwynn:
Thanks for the patch, @mastahyeti.

PR #713, opened 2016-01-26
Deprecate "revoke all application authorizations" from Octokit

tarebyte:
Since that endpoint [no longer exists](https://developer.github.com/changes/2014-04-08-reset-api-tokens/), it probably shouldn't be listed in Octokit.

### DIFF
--- a/lib/octokit/client/authorizations.rb
+++ b/lib/octokit/client/authorizations.rb
@@ -203,30 +203,6 @@ module Octokit
       end
       alias :delete_application_authorization :revoke_application_authorization
 
-      # Revoke all tokens for an app
-      #
-      # Applications can revoke all of their tokens in a single request
-      #
-      # @return [Boolean] Result
-      # @see https://developer.github.com/v3/oauth_authorizations/#revoke-all-authorizations-for-an-application
-      # @example
-      #  client = Octokit::Client.new(:client_id => 'abcdefg12345', :client_secret => 'secret')
-      #  client.revoke_all_application_authorizations
-      def revoke_all_application_authorizations(options = {})
-        opts = options.dup
-        key    = opts.delete(:client_id)     || client_id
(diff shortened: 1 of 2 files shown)


gjtorikian:
Nice. :+1:

lib/octokit/client/authorizations.rb
```diff
@@ -203,30 +203,6 @@ def revoke_application_authorization(token, options = {})
...
-        opts = options.dup
-        key    = opts.delete(:client_id)     || client_id
-        secret = opts.delete(:client_secret) || client_secret
-
-        as_app(key, secret) do |app_client|
-          app_client.delete "/applications/#{client_id}/tokens", opts
-
-          app_client.last_response.status == 204
-        end
-      rescue Octokit::NotFound
```
pengwynn: Removing a method will require a major version bump. I recommend we keep this method, [mark it as deprecated](https://github.com/octokit/octokit.rb/blob/28a594ff371ff3faf8ec50dd25bd65bd779ff43f/lib/octokit/client/downloads.rb#L13), add a [deprecation warning](https://github.com/octokit/octokit.rb/blob/28a594ff371ff3faf8ec50dd25bd65bd779ff43f/lib/octokit/warnable.rb#L12), and return `false`.

lib/octokit/client/authorizations.rb
```diff
@@ -203,30 +203,6 @@ def revoke_application_authorization(token, options = {})
...
-        false
-      end
-      alias :delete_application_authorization :revoke_application_authorization
```
pengwynn: Just noticed these aliases are incorrect dupes so we can :fire: them.

tarebyte:
@pengwynn I updated the PR based on your suggestions, let me know if there is anything else you would like to see changed.

Thanks!

pengwynn:
:cake:

PR #717, opened 2016-02-04
Update README with query example

jandudulski:
Add simple example how to pass query params to `get` method.

[Review on Reviewable](https://reviewable.io/reviews/octokit/octokit.rb/717)


pengwynn:
:cake:

Issue #721, opened 2016-02-27
Is there a method to import a Github repo?


hackerkid:
https://developer.github.com/v3/migration/source_imports/

Is this supported yet ?

pengwynn:
Methods like `Octokit::Client#start_import` do not yet exist (patches welcome :bowtie:), but you can drop down and use the lower-level HTTP interface. Something like:

```ruby
options = {
  :vcs => "subversion",
  :vcs_url => "http://svn.mycompany.com/svn/myproject",
  :vcs_username => "octocat",
  :vcs_password => "secret",
  :headers => {:accept => "application/vnd.github.barred-rock-preview"}
}
client = Octokit::Client.new :access_token => "yours"
client.put "https://api.github.com/repos/hackerkid/myrepo/import", options
```

hackerkid:
Thanks. I have added this to my todo list. I will send a PR when I complete :smile:

gjtorikian:
Head's up to @spraints who I also strong armed into trying to implement this. :grin:

PR #722, opened 2016-02-27
Add Source Imports preview to Octokit

tarebyte:
I saw that this was public, and it hadn't been added yet.

This adds the new Source Imports preview https://developer.github.com/v3/migration/source_imports

Feedback always wanted and appreciated.

/cc @pengwynn @spraints


lib/octokit/preview.rb
```diff
@@ -6,7 +6,8 @@ module Preview
     PREVIEW_TYPES = {
       :branch_protection => 'application/vnd.github.loki-preview+json'.freeze,
       :migrations        => 'application/vnd.github.wyandotte-preview+json'.freeze,
-      :licenses          => 'application/vnd.github.drax-preview+json'.freeze
+      :licenses          => 'application/vnd.github.drax-preview+json'.freeze,
+      :source_imports    => 'application/vnd.github.barred-rock-preview'.freeze
```
pengwynn: A trailing comma today will save us a diff tomorrow.

lib/octokit/client/source_import.rb
```diff
@@ -0,0 +1,90 @@
...
+      #   @client.source_import_progress("octokit/octokit.rb")
+      def source_import_progress(repo, options = {})
+        options = ensure_api_media_type(:source_imports, options)
+        get "#{Repository.path repo}/import", options
+      end
+
+      # List source import commit authors
+      #
+      # @param repo [Integer, String, Hash, Repository] A GitHub repository.
+      # @return [Array<Sawyer::Resource>] Array of hashes representing commit_authors.
```
pengwynn: Should we add `@see https://developer.github.com/v3/migration/source_imports/#get-import-progress` here?

pengwynn: It looks like there's a [`since` option](https://developer.github.com/v3/migration/source_imports/#parameters-1) in the hash, too.

tarebyte: @pengwynn is this enough for the comment?

```ruby
# @option options [String] :since Only authors found after this id are returned.
```

lib/octokit/client/source_import.rb
```diff
@@ -0,0 +1,91 @@
...
+      # @option values [String] :name The new Git author name.
+      # @return [Sawyer::Resource] Hash representing the updated commit author
+      # @see https://developer.github.com/v3/migration/source_imports/#map-a-commit-author
+      #
+      # @example
+      #   @client.map_source_import_commit_author("octokit/octokit.rb", 1, {
+      #     :email => "hubot@github.com",
+      #     :name => "Hubot the Robot"
+      #   })
+      def map_source_import_commit_author(repo, author, values, options = {})
```
pengwynn: Instead of building this with `repo` and the pretty esoteric `author_id` what if we leaned into hypermedia and have this method take a `url`. That you get from `source_import_commit_authors`? I'm guessing that's the only way you'd know about those authors anyway.

lib/octokit/client/source_import.rb
pengwynn: Yup!

pengwynn:
Looking good @tarebyte! I left a few comments.

/cc @hackerkid who asked for this in #721.

lib/octokit/client/source_import.rb
pengwynn: Looks like the diff moved as I hit update and lost my code example. Something like:

```ruby
def map_source_import_commit_author(author_url, values, options = {})
```

tarebyte: @pengwynn that make sense to me, do you know of an open source project I can run the import tests on?

If I import a git project from GitHub like I am doing in my tests there aren't any authors to pull the url from.

tarebyte: nvm I found a google code project that will work

tarebyte:
@pengwynn thank you for the review :sparkles: 

If everything looks good to you, I think this is ready to go!

hackerkid:
@tarebyte Thanks for adding source imports.  :sparkles: I was looking for this.

spec/octokit/client/source_import_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  describe ".start_source_import", :vcr do
+    it "starts a source import" do
+      assert_requested :put, github_url("/repos/#{@repo.full_name}/import")
+    end
+  end # .start_source_import
+
+  describe ".source_import_progress", :vcr do
+    it "returns the progress of the source import" do
+      @client.source_import_progress(@repo.full_name)
```
pengwynn: Might be smart to assert the response here:

```
expect(result).to be_kind_of Sawyer::Resource
```

spec/octokit/client/source_import_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+
+      expect(commit_author.email).to eql("hubot@github.com")
+      expect(commit_author.name).to eql("Hubot the Robot")
+      assert_requested :patch, commit_author_url
+    end
+  end # .map_source_import_commit_author
+
+  describe ".cancel_source_import", :vcr do
+    it "cancels the source import" do
+      @client.cancel_source_import(@repo.full_name)
```
pengwynn: And here:

```
expect(result).to be_true
```

pengwynn:
@tarebyte two small suggestions this time around, then I think we're good to merge.

tarebyte:
@pengwynn updated the specs with your suggestions :smile:

pengwynn:
:cake:

spraints:
:heart: Thanks for this!

PR #723, opened 2016-02-28
Update issue specs to use newly created repository

tarebyte:
I was working on https://developer.github.com/v3/issues/#lock-an-issue, but thought that this change was big enough that it deserved it's own PR before going on.

After deleting the cassettes I ran into an abuse limiting error because the tests were running too many operations on a single issue in a short span of time.

This updates the specs to create a new repository for each test, and create a new issue for that repository.

/cc @pengwynn


tarebyte:
Protip :tm: do not name your test org and your test user the same with only a one letter change on the end.

Example my test user is `tatooine-moisture-farmer` and my test org is `tatooine-moisture-farmerz` VCR got confused and used `<GITHUB_LOGIN>z` which broke the test running on CI.

Once I created a new org called `thorin-oakenshield` everything ran much better.

tarebyte:
@pengwynn a gentle bump, if you get a chance would you mind reading this over to see if it's good to merge?

joeyw:
I made some changes to the nesting, issue_timeline was in the wrong scope. I also changed an-repo to use the repo helper along with a time stamp. 

Some funny things happen when creating and deleting repos in quick succession, which is why I believe this was shot down in the past, but I don't see this really being abused so I added it. 

Since I merged the lock/unlock issue pull request there were some conflicts, but fixed that also in the merge.

PR #724, opened 2016-03-02
Docs for target_url & description status options

olivierlacan:
I was playing around with this neat API earlier and noticed that some really cool options 
were not documented on the Octokit side: 
- https://developer.github.com/v3/repos/deployments/#create-a-deployment-status

I can confirm that `target_url` and `description` work from testing them. The first 
adds a link to the status text in a PR pointing to the `target_url`. The second 
becomes the `title` attribute on that link.

This:
```ruby
client.create_deployment_status(url, "success", { target_url: "https://www.codeschool.com", description: "Things happened.!" })
```

Turns into that:
image


pengwynn:
Good catch. :fishing_pole_and_fish:

PR #725, opened 2016-03-04
Loosen sawyer dependency

nlopes:
This loosens sawyer addressable dependency which makes octokit less
strict on the addressable version.

### DIFF
--- a/octokit.gemspec
+++ b/octokit.gemspec
@@ -5,7 +5,7 @@ require 'octokit/version'
 
 Gem::Specification.new do |spec|
   spec.add_development_dependency 'bundler', '~> 1.0'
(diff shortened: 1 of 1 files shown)


gjtorikian:
It might be better to just set it as `~> 0.6`.

Could you expand on the issue with `addressable`? Is there a conflict somewhere?

nlopes:
We use octokit on a project that also has a direct dependency on addressable, but we are at a later release of addressable on that project. When adding octokit, bundle throws a tantrum about addressable needing to be at 2.5 but octokit requiring sawyer at 0.6 makes addressable be at 2.3.5.

pengwynn:
Thanks. :cake:

PR #733, opened 2016-03-16
Support getting milestones using a repository id

stmllr:
Milestones of a repository can be queried using the repository id.


stmllr:
If you like to have tests for this, please let me know. I left them out intentionally because
 * I did not find any other spec which uses the repository id for queries.
 * Repository.path is covered itself within `repository_spec.rb`
 * using repository id is not documented in GitHub API

tarebyte:
@stmllr it would be nice to have a few tests, just to ensure we don't add a regression in later.

Here is an example for you https://github.com/octokit/octokit.rb/blob/3dcb9184cba56f591d7b278dfd3c2db04de95a05/spec/octokit/client/repositories_spec.rb#L50-L54

stmllr:
@tarebyte Here we go...

stmllr:
any chance this PR gets merged?

pengwynn:
Nice. Thanks, @stmllr.

PR #738, opened 2016-03-28
Fix source_imports documentation URL

rmacklin:
https://developer.github.com/v3/repos/source_imports gives:
> Whoops, looks like that page doesn't exist.

It seems the URL has changed to https://developer.github.com/v3/migration/source_imports.


tarebyte:
Thanks @rmacklin!

PR #739, opened 2016-04-03
Escape options values in authorize_url

shvaikalesh:
`authorize_url` appends entries from `options` hash as query params *without* escaping values. This PR adds `CGI.escape` call to fix that, because `redirect_uri` and `scope` options probably have chars that should be encoded.

### DIFF
--- a/spec/helper.rb
+++ b/spec/helper.rb
@@ -2,7 +2,7 @@ if RUBY_ENGINE == 'ruby'
   require 'simplecov'
   require 'coveralls'
 
-  SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter[
+  SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new [
     SimpleCov::Formatter::HTMLFormatter,
     Coveralls::SimpleCov::Formatter
   ]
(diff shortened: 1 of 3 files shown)


spec/helper.rb
```diff
@@ -2,7 +2,7 @@
   require 'simplecov'
   require 'coveralls'
 
-  SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter[
+  SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new [
```
joeyw: 🍰

spec/helper.rb
```diff
@@ -13,7 +13,7 @@
 require 'octokit'
 require 'rspec'
 require 'webmock/rspec'
-require "base64"
+require 'base64'
```
joeyw: ✨

joeyw:
Hi @shvaikalesh apologies for the slow response. In your process of fixing a bug, the code is much nicer now also. Thank you 👍

shvaikalesh:
@joeyw, thanks for the review. If you will publish new version to rubygems, you may also want to adjust [required ruby version](https://github.com/octokit/octokit.rb/blob/master/octokit.gemspec#L18) (2.0) to match `README.md`.

joeyw:
Thanks, I made #753 so we don't forget 👍

PR #742, opened 2016-04-14
Fix Default.middleware documentation comment

justCxx:


lib/octokit/default.rb
```diff
@@ -106,7 +106,7 @@ def login
 
       # Default middleware stack for Faraday::Connection
       # from {MIDDLEWARE}
-      # @return [String]
+      # @return [RACK_BUILDER_CLASS]
```
joeyw: Good catch @justCxx. 

Since you pointed this out, I reviewed `RACK_BUILDER_CLASS` to see the actual class returned. I think this would be more accurate if it was `@return [Faraday::Builder or Faraday::RackBuilder]` which are the classes that could be returned. Do you mind making the change? 

Thanks for taking the time to report this. 👍

joeyw:
Sweet 👍

PR #748, opened 2016-05-03
Include new methods for Source Import API enhancements 

LizzHale:
New methods were introduced for the Source Import API to support importing large files with Git LFS. This PR updates octokit to reflect those changes. 

* [x] Updates #start_source_import to optionally take vcs in option hash
* [x] Supports getting list of large files for source import
* [x] Supports updating or restarting a source import 

/cc @pengwynn @tarebyte @gjtorikian @jonmagic @spraints


lib/octokit/client/source_import.rb
```diff
@@ -41,6 +41,27 @@ def source_import_progress(repo, options = {})
...
+      # Update source import with authentication/project_choice
+      # Restart source import if no options are passed
+      #
+      # @param repo [Integer, String, Hash, Repository] A GitHub repository.
+      # @return [Sawyer::Resource] Hash representing the repository import
+      # @see https://developer.github.com/v3/migration/source_imports/#update-existing-import
+      # @option options [String] :vcs_username If authentication is required, the username to provide to vcs_url.
+      # @option options [String] :vcs_password If authentication is required, the password to provide to vcs_url.
+      # @option options [String] :vcs The originating VCS type. Can be one of "subversion", "git", "mercurial", or "tfvc".
+      # @option options [String] :tfvc_project For a tfvc import, the name of the project that is being imported.
```
spraints: Instead of listing vcs and tfvcproject, can this refer to the project choices list on the get endpoint? The exact attributes should be one of the project choices that the API returned.

LizzHale: 👍

lib/octokit/client/source_import.rb
```diff
@@ -9,9 +9,9 @@ module SourceImport
       # Start a source import to a GitHub repository using GitHub Importer.
       #
       # @param repo [Integer, String, Hash, Repository] A GitHub repository.
-      # @param vcs [String] The originating VCS type. Can be one of "subversion", "git", "mercurial", or "tfvc".
       # @param vcs_url [String] The URL of the originating repository.
       # @param options [Hash]
+      # @option options [String] :vcs The originating VCS type. Can be one of "subversion", "git", "mercurial", or "tfvc".
```
LizzHale: The Source Imports API no longer requires the `vcs` attribute to start an import. I wanted to bring particular attention to this change as it breaks the current implementation by moving :vcs to the options hash.

tarebyte: @LizzHale I think that's fine, this users know that is for a preview.

Would you mind also updating the example further below to further clarify how the method should work?

lib/octokit/client/source_import.rb
```diff
@@ -88,6 +109,35 @@ def cancel_source_import(repo, options = {})
         options = ensure_api_media_type(:source_imports, options)
         boolean_from_response :delete, "#{Repository.path repo}/import", options
       end
+
+      # List source import large files
+      #
+      # @param repo [Integer, String, Hash, Repository] A GitHub repository.
+      # @param options [Hash]
+      # @option options [Fixnum] :page Page of paginated results
```
tarebyte: Is this option something other than normal GitHub pagination?

If it isn't we should ✂️  this line since we don't talk about pagination in any of the other methods.

tarebyte: Yup just answered my own question be reading below, carry on.

tarebyte:
@LizzHale this looks 👍 to me, if you recreate the cassettes for the source import I have a feeling all of your tests will pass.

Give me a shout if I can be of anymore help!

lib/octokit/client/source_import.rb
LizzHale: >Would you mind also updating the example further below to further clarify how the method should work?

Good catch! Will do.

lib/octokit/client/source_import.rb
LizzHale: Updated here a139ee6

LizzHale:
> if you recreate the cassettes for the source import I have a feeling all of your tests will pass.

Thanks for the feedback @tarebyte. I recreated the test cassettes but I'm still getting CI failures on unrelated tests. Am I missing something?

joeyw:
@LizzHale Looks like an issue unrelated to your tests or cassettes. I pulled your branch and the tests pass, so it's good from that point.

LizzHale:
Thanks @joeyw

lib/octokit/client/source_import.rb
```diff
@@ -9,22 +9,23 @@ module SourceImport
...
+      # @option options [String] :vcs The originating VCS type. Can be one of "subversion", "git", "mercurial", or "tfvc".
       # @option options [String] :vcs_username If authentication is required, the username to provide to vcs_url.
       # @option options [String] :vcs_password If authentication is required, the password to provide to vcs_url.
       # @option options [String] :tfvc_project For a tfvc import, the name of the project that is being imported.
       # @return [Sawyer::Resource] Hash representing the repository import
       # @see https://developer.github.com/v3/migration/source_imports/#start-an-import
       #
       # @example
-      #   @client.import("octokit/octokit.rb", "subversion", "http://svn.mycompany.com/svn/myproject" {
+      #   @client.import("octokit/octokit.rb", "http://svn.mycompany.com/svn/myproject" {
```
joeyw: Missing a comma here in the example call:

```ruby
"http://svn.mycompany.com/svn/myproject" {
```
```ruby
"http://svn.mycompany.com/svn/myproject", {
```

LizzHale: Thanks -- I also updated the method name in the example while I was there

lib/octokit/client/source_import.rb
joeyw: > @LizzHale I think that's fine, this users know that is for a preview.

I agree with the point that users accept additional conditions when using preview APIs, but I would like to see that apply to the API itself only. 

In this case if someone is using the preview API raw, there code wont break because of it no longer being required, however that is not the case for this change in Octokit. Since wrappers are supposed to make things easier, I would like to see us gracefully deprecate this even for a preview API. I believe we should stick to our versioning for Octokit and not break APIs between nonmajor versions. We can promote the new API while supporting the existing deprecated one, then remove the deprecated call in the next major version.

I believe something along these lines would be fine? What do you all think?

```ruby
def start_source_import(repo, vcs_url, options = {}, _deprecated_args=nil)
  if _deprecated_args
    vcs = vcs_url
    vcs_url = options
    options = _deprecated_args.merge(:vcs => vcs)
  end

  options = ensure_api_media_type(:source_imports, options.merge(:vcs_url => vcs_url))
  put "#{Repository.path repo}/import", options
end
```

LizzHale: @joeyw I really like the idea of gracefully deprecating this and that looks like a great way to do it. I'll play around with the method you provided and see if I can think of a way to cover the case when no options are passed - for example:

`@client.start_source_import(@repo.full_name, "svn", "https://rphotogallery.googlecode.com/svn/")`

joeyw: @LizzHale We've done some splat action in the past for situations like this, I'm not a huge fan of the splat operator personally. I would much rather see a type check vs hash on the options parameter.

LizzHale:
I had some time today to work on gracefully deprecating requiring the vcs parameter in the start_source_import method in commit ad10199. This should handle method calls with the following formats:

start_source_import(repo, vcs, vcs_url)
start_source_import(repo, vcs, vcs_url, options)
start_source_import(repo, vcs_url)
start_source_import(repo, vcs_url, options)

To test the deprecation I split the tests into two new describe blocks. The pre-deprecation block  asserts that `start_source_import(repo, vcs, vcs_url)` raises the deprecation warning. All the other tests were left untouched and nested in a post-deprecation block. 

Let me know what you think @joeyw.

LizzHale:
Oops... looks like the test repos weren't being deleted at the end each test. I reverted and committed only the change to the method itself and will try to find a more graceful way to test the deprecation warning without all the nested describes.

LizzHale:
I hit a blocker today. It appears that the repository (https://rphotogallery.googlecode.com/svn/) that we're using to import in the specs is failing so I'm getting into a loop [here](https://github.com/LizzHale/octokit.rb/blob/081dd33cbe68fce645e09d3f57fdabf2e690bf9d/spec/octokit/client/source_import_spec.rb#L54) when trying to create new cassettes. I'm assuming this is new behavior since I didn't run into it previously. @tarebyte have you seen something like this in the past?

tarebyte:
@LizzHale I haven't hit that problem before, is there another SVN repository we can use?

joeyw:
@LizzHale https://vsarbranchingguide.svn.codeplex.com/svn ? :)

spraints:
Codeplex svn doesn't work in our importer because it doesn't support svnrdump. Does the test repo need to be an svn repository? Or would an hg or git repository work?

LizzHale:
@joeyw, unfortunately imports from Codeplex's SVN bridge are not supported because Codeplex does not support svnrdump. :( It's also possible to connect to Codeplex using TFS but I just learned you need to be a  team members (so likely need to authenticate). 

I have a few svn repos I've used but they all require auth so that's not ideal. I'll keep looking though, I'm sure we'll find one. I appreciate the help @joeyw and @tarebyte

LizzHale:
Oh @spraints beat me to it! :laughing:

joeyw:
@LizzHale This doesn't need to be an svn import since the intention isnt to test the api, only that we can call it from octokit, we can import https://github.com/octokit/octokit.rb or other possibly lighter weight repo.

LizzHale:
@spraints am I correct in assuming that importing a git repo won't create a list of authors? If so, we'll need a non-git repo since the existing .map_source_import_commit_author test needs a repo that returns at least 1 author. But other than that, I agree we're free to use whatever VCS.

spraints:
> am I correct in assuming that importing a git repo won't create a list of authors? If so, we'll need a non-git repo since the existing .map_source_import_commit_author test needs a repo that returns at least 1 author.

Right, importing from Git doesn't not generate an author list. Would a Mercurial repository on bitbucket work? It'd probably also be good to make the upstream source URL a variable that the tester needs to supply in order to regenerate fixtures (similar to how tokens are handled), since the upstream source code will be outside of octokit's control.

LizzHale:
@spraints was kind enough to provide a repo url so I could get my tests generated and back on track.

>It'd probably also be good to make the upstream source URL a variable that the tester needs to supply in order to regenerate fixtures (similar to how tokens are handled), since the upstream source code will be outside of octokit's control.

I'm supportive of making the test repo url a variable. What are your opinions @joeyw @tarebyte? I think the tests are setup in such a way that as long as there's one author any subversion, mercurial, or team foundation repository would work. 

Aside from that, this is ready for :eyes:

lib/octokit/client/source_import.rb
```diff
@@ -9,22 +9,34 @@ module SourceImport
...
       # @example
-      #   @client.import("octokit/octokit.rb", "subversion", "http://svn.mycompany.com/svn/myproject" {
+      #   @client.start_source_import("octokit/octokit.rb", "http://svn.mycompany.com/svn/myproject", {
+      #    :vcs           => "subversion",
       #    :vcs_username" => "octocat",
       #    :vcs_password  => "secret"
       #   })
-      def start_source_import(repo, vcs, vcs_url, options = {})
-        options = ensure_api_media_type(:source_imports, options.merge(:vcs => vcs, :vcs_url => vcs_url))
+      def start_source_import(repo, vcs_url, options = {}, _deprecated_args=nil)
```
pengwynn: @LizzHale would you mind tackling this with a splat for consistency with the rest of the library? You can use the  `@overload` yardoc tag to communicate the different approaches. 

Here's some [prior art](https://github.com/octokit/octokit.rb/blob/719483524b930146580a997284fdd9719a5b0cdf/lib/octokit/client/commits.rb#L13-L31). I don't think your tests will change.

LizzHale: > would you mind tackling this with a splat for consistency with the rest of the library?

Happy to -- I'll carve out some time this week to get that in.

LizzHale: @pengwynn: updated #start_source_import to use a splat here 4048720

pengwynn:
🍰

PR #760, opened 2016-05-13
Add Reactions Preview to Octokit

tarebyte:
This PR adds the Reactions Preview API to Octokit https://developer.github.com/v3/reactions/.

Currently having issues creating a commit through the API, in order to there to be a Pull Request.

Feedback, welcome!

/cc @pengwynn for review


lib/octokit/client/reactions.rb
```diff
@@ -0,0 +1,154 @@
...
+      def commit_comment_reactions(repo, id, options = {})
+        options = ensure_api_media_type(:reactions, options)
+        get "#{Repository.path repo}/comments/#{id}/reactions", options
+      end
+
+      # Create a reaction for a commit comment
+      #
+      # @param repo [String, Hash, Repository] A GitHub repository
+      # @param id [Integer] The id of the commit comment
+      # @param reaction [Integer] The Reaction
```
joeyw: `reaction` is a string parameter

lib/octokit/client/reactions.rb
```diff
@@ -0,0 +1,154 @@
...
+      # @see https://developer.github.com/v3/reactions/#create-reaction-for-a-commit-comment
```
joeyw: Since the api doesn't have a method yet for listing the possible reactions. We should add a `@see` link to https://developer.github.com/v3/reactions/#reaction-types

lib/octokit/client/reactions.rb
```diff
@@ -0,0 +1,154 @@
...
+      #
+      # @example
+      #   @client.create_commit_comment_reactions("octokit/octokit.rb", 1)
```
joeyw: This example is missing the required reaction parameter.

stmllr:
The repo argument also accepts Integer. However this is missing in the type hint comment:

``` diff
- # @param repo [String, Hash, Repository] A GitHub repository
+ # @param repo [Integer, String, Hash, Repository] A GitHub repository
```

tarebyte:
@stmllr updated in 48864e7

kneemer:
@tarebyte, just wanted to let you know that the Reactions API has been updated to return the full user object instead of just `user_id` https://developer.github.com/changes/2016-06-07-reactions-api-update/

PR #761, opened 2016-05-17
Adding Octokit::AccountSuspended for when the 403 is due to a suspend…

koglinjg:
…ed account


koglinjg:
Created this specific 403 exception class so that I can catch account suspended errors and not have to rely on catching the generic Octokit::Forbidden

pengwynn:
Nice. 🍰

PR #763, opened 2016-05-25
Update add_collaborator and add_team_repository docs to include permission option

rmacklin:
I thought it would be useful to document this option.

If you regenerate the docs with this change, it'll look like:

screen shot 2016-05-25 at 1 48 19 pm

Previously it looked like:

screen shot 2016-05-25 at 1 48 48 pm


pengwynn:
✨

PR #770, opened 2016-06-23
Add `Client.lock_issue` and `Client.unlock_issue`

davidcelis:
With the recent finalizing of the API endpoints to lock and unlock
issues, we should include that functionality in Octokit.rb. This patch
adds two methods to lock or unlock an issue. Because the corresponding
API endpoints return a 204 NO CONTENT on success, we just coerce the
response into a Boolean for the Octokit methods.

Side note: not entirely sure what to do to get VCR to work in this repo. I
wrote some corresponding specs but I'm waiting to see what CI says.

### DIFF
--- a/spec/octokit/client/issues_spec.rb
+++ b/spec/octokit/client/issues_spec.rb
@@ -113,6 +113,20 @@ describe Octokit::Client::Issues do
       end
     end # .reopen_issue
 
+    describe ".lock_issue" do
+      it "locks an issue" do
+        @client.lock_issue(@test_repo, @issue)
+        assert_requested :put, github_url("/repos/#{@test_repo}/issues/#{@issue.number}/lock")
+      end
+    end # .lock_issue
+
+    describe ".unlock_issue" do
+      it "unlocks an issue" do
+        @client.unlock_issue(@test_repo, @issue)
+        assert_requested :delete, github_url("/repos/#{@test_repo}/issues/#{@issue.number}/lock")
+      end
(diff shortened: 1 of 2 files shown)


spec/octokit/client/issues_spec.rb
```diff
@@ -113,6 +113,20 @@
       end
     end # .reopen_issue
 
+    describe ".lock_issue" do
+      it "locks an issue" do
+        @client.lock_issue(@test_repo, @issue)
+        assert_requested :put, github_url("/repos/#{@test_repo}/issues/#{@issue.number}/lock")
+      end
+    end # .close_issue
```
pengwynn: Copypasta.

spec/octokit/client/issues_spec.rb
```diff
@@ -113,6 +113,20 @@
...
+
+    describe ".unlock_issue" do
+      it "unlocks an issue" do
+        @client.unlock_issue(@test_repo, @issue)
+        assert_requested :delete, github_url("/repos/#{@test_repo}/issues/#{@issue.number}/lock")
+      end
+    end # .reopen_issue
```
pengwynn: And here.

pengwynn:
> Side note: not entirely sure what to do to get VCR to work in this repo

@davidcelis The README [has your back](https://github.com/octokit/octokit.rb#running-and-writing-new-tests). You need to run the tests locally and check in your new fixtures because we don't record in CI.

PR #777, opened 2016-07-02
Repository: Avoid obsolete warning using URI::ABS_URI

olleolleolle:
Try to avoid the warning "URI.regexp is obsolete" by using one of the specific regular expressions made available to the `URI` module.

Benefit: much less chatty output. 

Example: [this test output](https://travis-ci.org/skywinder/github-changelog-generator/jobs/165365731)


joeyw:
Thanks! I also took out the `\A and \z` as the regex is slightly different and now includes it.

PR #780, opened 2016-07-15
Dupe the options hash in Repositories#create_repo

davidcpell:
WHY:
I was getting unexpected behavior from my program in a class that wraps
Octokit. After looking at the source code I realized it was because the
hash I was passing to Repositories#create_repo was being mutated inside
the method. I'm thinking it would be better to dupe the hash before
mutating it to avoid surprises to the user. I named the duped version `opts` so that it would be clear inside the method that this object is not the same as the one passed in.


davidcpell:
any thoughts on this @pengwynn @sferik?

joeyw:
Good catch, Ive opened #813 to remove this unexpected behavior from the rest of the library.

Issue #779, opened 2016-07-08
Updated support for Protected Branches


jwhitcraft:
Since the Protected Branches API updated, it would be nice to have it make it's way into octokit.

https://developer.github.com/changes/2016-06-27-protected-branches-api-update/

PR #784, opened 2016-07-28
Update protected branches API

atmos:
I'm trying to use the new protected branches API. I think I did this right.

Fixes https://github.com/octokit/octokit.rb/issues/779


pengwynn:
Thanks @atmos. Looks good on my end. I'll give @tma and @arthurschreiber a chance to take a look before I merge in the morning.

arthurschreiber:
I think this looks fine. 👍

PR #786, opened 2016-08-11
adding support for getting a single deployment by id

ghostsquad:
This PR is to add support for getting a single deployment by id. Unfortunately, I'm not able to get my test to work because of a credential issue:

```
Failures:

  1) Octokit::Client::Deployments.deployment gets single deployment
     Failure/Error: deployment = @client.deployment(@test_repo, 137)
     Octokit::Unauthorized:
       GET https://api.github.com/repos/api-padawan/api-sandbox?auto_init=true: 401 - Bad credentials // See: https://developer.github.com/v3
```

Let me know how I can resolve this, and I'll make the necessary changes.

### DIFF
--- a/lib/octokit/client/deployments.rb
+++ b/lib/octokit/client/deployments.rb
@@ -16,6 +16,11 @@ module Octokit
       end
       alias :list_deployments :deployments
 
+      def deployment(repo, id, options = {})
+        get("#{Repository.path repo}/deployments/#{id}", options)
+      end
+      alias :get_deployment :deployment
+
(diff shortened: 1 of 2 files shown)


ghostsquad:
Looks like I need to finish reading the readme to solve this problem 👍

joeyw:
Theres not an endpoint for this currently, but there are options for the https://developer.github.com/v3/repos/deployments/#list-deployments endpoint, which you can pass the SHA to get a specific deployment.

PR #797, opened 2016-09-07
#upload_assets should open files in read-only mode

trevorrowe:
Calling `#upload_assets` on a file caused it to be opened using "r+b" which is read-write. This prevents `#upload_file` from working with assets that are read-only. Given the method does not attempt to modify the file, the correct open flag should be "rb".


joeyw:
Thanks! ⭐️

PR #803, opened 2016-09-30
Allow manual concat for auto paginated resources

sclinede:
I'm working under open-source analytic tool and need to process huge projects via octokit. Auto pagination feature works great and even has feature to concat results manually.
But we cannot use it right from resource request.
I was forced to write dirty hook over pagination method to use it. I think it make sence to provide such feature without additional movements in side code. It can also save memory a lot in case of big projects.
Now we can use it like that:

``` ruby
Octokit.auto_paginate = true

open_issues = []
first_page_data = Octokit.issues 'rails/rails' do |_, next_page|
  next_page.data.each do |issue|
    open_issues << issue if issue[:state] == 'open'
  end
end
first_page_data.each do |issue|
  open_issues << issue if issue[:state] == 'open'
end

open_issues.length

```


sclinede:
Please @pengwynn. 

 I'll rebase on approval.

sclinede:
@tarebyte,@pengwynn :arrow_up:

sclinede:
Confilicts are fixed. Please, @tarebyte,@pengwynn

sclinede:
:up: @tarebyte,@pengwynn

pengwynn:
@sclinede oops, I'm still seeing conflicts. (Perhaps due to merges I just made to master?) I'll pull this down locally and try to sort it out unless you beat me to it.

sclinede:
I'm fighting with floating tests fails... It will take some time @pengwynn

sclinede:
Aaaand. Done :package: @pengwynn

pengwynn:
@sclinede whoa, large diff is large! 😄 Are the cassette updates required for the code changes or are those formatting tweaks only?

I was kicking the tires on this locally and I'm not sure it's behaving as described in the updated docs. It seems the `data` parameter passed to the block is always the first page?

Given the [performance problems with declared block parameters in Ruby](https://docs.omniref.com/ruby/2.2.0/symbols/Proc/yield#annotation=4087638&line=711&hn=1), I wonder if there is a way to provide the behavior with `block_given?` and `yield` that doesn't incur that overhead. 🤔

sclinede:
@pengwynn 
1. Yes cassettes are required. Pagination works using special header, so they should be introduced in recorded requests.
2. Yes. Due to ```paginate``` implementation [here](https://github.com/octokit/octokit.rb/blob/master/lib/octokit/connection.rb#L78) we receive in a block two variables: ```data``` - Sawyer::Response#data which could be an Array and behave like accumulator and ``` last_response ``` is Sawyer::Response.  Also ``` data ``` in that case contains only data of the first request. If request don't have next page then block will not be yielded and we will have only data that method call returned. 
And method returns first Sawyer::Response#data that was gathered.
So. I've added to docs real usage of that case. We can use this one if you'd prefer:
```ruby
Octokit.auto_paginate = true

open_issues = []
accumulated_data = Octokit.issues 'rails/rails' do |_, next_page|
  next_page.data.each do |issue|
    open_issues << issue if issue[:state] == 'open'
  end
end
accumulated_data.each do |issue|
  open_issues << issue if issue[:state] == 'open'
end

open_issues.length
```

But. Maybe it is better to introduce separate method for that and just remove all other complexity.
Something like that:
```ruby

    def manual_pages_processing(url, options = {})
      raise ArgumentError, 'Auto pagination expceted to be turned on' unless @auto_paginate
      raise ArgumentError, 'Block expected for manual processing' unless block_given?

      opts = parse_query_and_convenience_headers(options.dup)
      opts[:query][:per_page] ||= @per_page || 100

      data = []
      request(:get, url, opts.dup)
      next_page = @last_response

      while next_page
        yield(data, next_page)

        return data unless @last_response.rels[:next] && rate_limit.remaining > 0
        next_page = @last_response.rels[:next].get(:headers => opts[:headers])
      end

      data
    end

``` 

3, Okay. I'll try to implement it using ```block_given?```

kytrinyx:
> Maybe it is better to introduce separate method for that and just remove all other complexity.
> [...]
> This way we'll not use it in every method but call it explicitly somehow.

@sclinede I had a chat with @tarebyte about it today as well, and we think that you're right about rethinking pagination from scratch and avoiding all this other complexity.

Would you like to open an issue to discuss what the optimal pagination pattern would look like?

Would you be willing to open an issue that documents how to use the pagination method using the block syntax in the pagination section of the README?

I'm going to go ahead and close this pull request, as we will not be pulling in the changes as in.

Thanks so much for taking the time to help think through this! 🌷

stmllr:
Please link potential new issues with this PR, so that we can follow-up through notifications. Thanks

PR #805, opened 2016-10-04
Return nil on timeout fetching stats

etiennebarrie:
Currently any method from `Octokit::Client::Stats` may return an `Array` of `Sawyer::Resource` objects, `nil` or an empty `Sawyer::Resource`. That's because returning `data` upon timeout will return the data of the 202 response, which is an empty JSON object.

The documentation for the private `get_stats` method doesn't mention returning a single `Sawyer::Resource` so I assumed it's not a specified behavior. Additionally when using `retry_timeout: nil`, it returns nil when it gets a 202.

This changes `get_stats` to return `nil` in this case too.


joeyw:
You are correct, it should work this way. Thanks!

PR #815, opened 2016-10-23
Add Pages.request_page_build

joeyw:
Add Pages.request_page_build for https://developer.github.com/v3/repos/pages/#request-a-page-build

Change Pages.pages to use preview media type to gain additional data


joeyw:
ps remember oceanic 815 ✈️

lib/octokit/client/pages.rb
```diff
@@ -12,7 +12,8 @@ module Pages
       # @return Sawyer::Resource A GitHub Pages resource
       # @see https://developer.github.com/v3/repos/pages/#get-information-about-a-pages-site
       def pages(repo, options = {})
-        get "#{Repository.path repo}/pages", options
+        opts = ensure_api_media_type(:pages, options)
+        get "#{Repository.path repo}/pages", opts
```
parkr: @joeyw Per the documentation you listed, this is actually incorrect. The preview header is only required for `POST`, not `GET.

PR #819, opened 2016-10-23
Bugfix: Enable branch protection without enforcing required_status_checks.

mgreensmith:
A small bugfix for the `protect_branch` implementation:

API docs say that you can pass `null` for either/both of `required_status_checks` and `restrictions`, but the existing client implementation sets non-nil default values for`required_status_checks`, which makes it impossible to unset.

This change allows branch protection to be enabled without enforcing `required_status_checks`.


joeyw:
Looks good, thanks for taking the time. ⭐️ 

Since that is optional now I would like to move it to the options parameter, but I have some other things I see I want to clean up unrelated to this specific change that ill put in a new pull.

👍

PR #820, opened 2016-10-23
Add Projects APIs

joeyw:
Add repositories projects endpoints

https://developer.github.com/v3/projects/
- [X] List projects
- [X] List organization projects
- [X] Get a project
- [X] Create a project
- [X] Create an organization project
- [X] Update a project
- [X] Delete a project
- [X] List columns
- [X] Get a column 
- [X] Create a column
- [X] Update a column 
- [X] Delete a column
- [x] Move a column
- [X] List projects cards
- [x] Get a project card
- [x] Create a project card
- [x] Update a project card
- [x] Delete a project card
- [x] Move a project card


joeyw:
This covers all the current Projects api end points now

lib/octokit/client/projects.rb
```diff
@@ -0,0 +1,314 @@
...
+      # List projects for a repository
+      #
+      # Requires authenticated client
+      #
+      # @param repo [Integer, String, Repository, Hash] A GitHub repository
+      # @return [Array<Sawyer::Resource>] Repository projects
+      # @see https://developer.github.com/v3/projects/#list-repository-projects 
+      # @example
+      #   @client.projects('octokit/octokit.rb')
+      def projects(repo, options = {})
```
bswinnerton: I wonder if it would be worthwhile to unify this method with [`org_projects`](https://github.com/octokit/octokit.rb/pull/820/files#diff-9607adc63ef94bd2878d32f654bfca81R52) based on whether or not the `repo` parameter has a slash in it. I'm not sure whether it the convenience would trump the explicitness, though. What do you think?

lib/octokit/connection.rb
```diff
@@ -160,7 +160,7 @@ def request(method, path, data, options = {})
     # @return [Boolean] True on success, false otherwise
     def boolean_from_response(method, path, options = {})
       request(method, path, options)
-      @last_response.status == 204
+      @last_response.status == 204 || @last_response.status == 201
```
bswinnerton: Ah, I see what you mean now regarding [your comment](https://platform.github.community/t/about-the-projects-rest-api-category/66/2?u=bswinnerton) in the Platform Community Discourse. I'll chime in on that thread shortly of our original intentions of returning a `201` here.

lib/octokit/client/projects.rb
joeyw: The repository endpoints can also take an id instead of login/name so I think it could get a bit messy quick.

lib/octokit/connection.rb
joeyw: I see what you mean. Do you think it would be better to return the response here even if it's empty now instead of a Boolean?

PR #821, opened 2016-10-23
Change #protect_branch and #unprotect_branch, cleanup their specs

joeyw:
# protect_branch `required_status_check` is no longer a required parameter and should be passed in the options hash
# unprotect_branch now returns a boolean

Specs for these are improved a be and reorganized also

/cc @pengwynn


pengwynn:
🍰

PR #822, opened 2016-10-24
Remove requested 'Accept' header from GET /repos/o/r/pages

parkr:
@joeyw This addresses my comment from https://github.com/octokit/octokit.rb/pull/815/files#r84807489.

This is currently broken in v4.4.0.


parkr:
Thank you, @pengwynn!

PR #830, opened 2016-11-09
Fixes #727 - quick-n-dirty monkey patch

pvdb:
@pengwynn, in https://github.com/octokit/octokit.rb/issues/727#issuecomment-259302307 you wrote:

> If you want to push up a monkey patch, that'd be great.

And here it is, in all its glory!  :smile:

Because it's a monkey patch, it doesn't necessarily look pretty, but it works and the tests are green... but before I spend more time making it look acceptable, I thought I'd run the approach by you to see if you're OK with it, or else if you'd like to suggest a different direction.

The proof of the pudding is in eating it, or so they say:

```ruby
[1] pry(main)> Octokit::VERSION
=> "4.6.0"
[2] pry(main)> Sawyer::VERSION
=> "0.8.0"
[3] pry(main)> Addressable::VERSION::STRING
=> "2.5.0"
[4] pry(main)> repo = Octokit::Client.new.repository('octokit/octokit.rb') ; nil
=> nil
[5] pry(main)> repo.rels[:git].href
=> "git://github.com/octokit/octokit.rb.git"
[6] pry(main)> repo.rels[:svn].href
=> "https://github.com/octokit/octokit.rb"
[7] pry(main)> repo.rels[:ssh].href
=> "ssh://git@github.com/octokit/octokit.rb.git"
[8] pry(main)> _
```

Note how `repo.rels[:ssh].href` is now a RFC 2396 compliant SSH URL, as opposed to a shorter, `scp`-like URL, which is why `addressable` is no longer throwing the exception documented in #727.

PR #833, opened 2016-11-14
Handle malformed SSH URL format with recent Addressable versions

### LINKED ISSUE #727: addressable v2.5.0 causing issue in octokit.rb v4.6.0
Hi peeps,

After installing the latest release version of `octokit.rb` - i.e. `v4.3.0` - I noticed that the following query:

``` ruby
Octokit::Client.new.repository('octokit/octokit.rb').rels[:ssh].href
```

... which used to work in earlier releases, and returned:

``` ruby
"git@github.com:octokit/octokit.rb.git"
```

... now throws this exception instead:

``` ruby
Addressable::URI::InvalidURIError: Invalid scheme format: git@github.com
from /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/addressable-2.4.0/lib/addressable/uri.rb:867:in `scheme='
```

This is demonstrated in a bit more detail in the below `pry` sessions... I haven't had a chance yet to dig into the `sawyer` or `addressable` changes to determine whether this new behaviour is a bug or a feature, but I thought I'd raise it here already for awareness, as it is definitely a regression issue.

Thanks,

@pvdb

---

**previous release: `octokit.rb v4.2.0` with `sawyer v0.6.0` and `addressable v2.3.8`**

``` ruby
[1] pry(main)> Octokit::VERSION
=> "4.2.0"
[2] pry(main)> Sawyer::VERSION
=> "0.6.0"
[3] pry(main)> Addressable::VERSION::STRING
=> "2.3.8"
[4] pry(main)> repo = Octokit::Client.new.repository('octokit/octokit.rb') ; nil
=> nil
[5] pry(main)> repo.rels[:git].href
=> "git://github.com/octokit/octokit.rb.git"
[6] pry(main)> repo.rels[:svn].href
=> "https://github.com/octokit/octokit.rb"
[7] pry(main)> repo.rels[:ssh].href
=> "git@github.com:octokit/octokit.rb.git"
[8] pry(main)> _
```

**current release: `octokit.rb v4.3.0` with `sawyer v0.7.0` and `addressable v2.4.0`**

``` ruby
[1] pry(main)> Octokit::VERSION
=> "4.3.0"
[2] pry(main)> Sawyer::VERSION
=> "0.7.0"
[3] pry(main)> Addressable::VERSION::STRING
=> "2.4.0"
[4] pry(main)> repo = Octokit::Client.new.repository('octokit/octokit.rb') ; nil
=> nil
[5] pry(main)> repo.rels[:git].href
=> "git://github.com/octokit/octokit.rb.git"
[6] pry(main)> repo.rels[:svn].href
=> "https://github.com/octokit/octokit.rb"
[7] pry(main)> repo.rels[:ssh].href
Addressable::URI::InvalidURIError: Invalid scheme format: git@github.com
from /usr/local/var/rbenv/versions/2.3.0/lib/ruby/gems/2.3.0/gems/addressable-2.4.0/lib/addressable/uri.rb:867:in `scheme='
[8] pry(main)> _
```

pengwynn:
An alternate approach to #830 to fix #727.


pvdb:
Hey @pengwynn,

Thanks for looking into this, and for taking the time to fix it!

You wrote:
> Ideally we'd fix this server-side, but I'm not sure of the impact to adding the `ssh://` scheme to these.

Yes, I agree with fixing it server-side, but I also agree that that would be a drastic and possibly quite disruptive change, as I can imagine a lot of people and things now assuming and being dependent on the `scp`-like convenience format for clone URLs _(returned not just by the GitHub API, BTW, but also by our friends over at Bitbucket and the likes)_.

As argued in #727, the GitHub API returning an RFC 2396 compliant URL would probably be a good thing, and `git` itself insisting on a proper SSH clone URL, instead of making allowances for the `scp`-like convenience format, would be even better, but I guess it's no use crying over spilt milk :smile:

I've re-ran my little testcase against your branch, and yes: it works... it no longer raises the exception, and now returns the `scp`-like URL it used to :thumbsup:

```ruby
[1] pry(main)> repo = Octokit::Client.new.repository('octokit/octokit.rb') ; nil
=> nil
[2] pry(main)> repo.rels[:git].href
=> "git://github.com/octokit/octokit.rb.git"
[3] pry(main)> repo.rels[:svn].href
=> "https://github.com/octokit/octokit.rb"
[4] pry(main)> repo.rels[:ssh].href
=> "git@github.com:octokit/octokit.rb.git"
[5] pry(main)> _
```

However... :smile:

AFAICT, the issue I discovered and reported in #727 does not point at a systemic and general issue with `Sawyer::Relation#href`, but rather at a very specific and localized issue with the `ssh_url` field returned by [the various repository endpoints in the GitHub API](https://developer.github.com/v3/repos/#get) and "consumed"/"processed" by `octokit.rb`.

In general, my preferred approach is to fix issues where they cause problems, and by doing so limiting the "radiation zone" of the change, hence my proposal in #830.

The concern I have about the changes in this PR _(even though it fixes the issue/symptom I reported, don't get me wrong)_ is the risk of unintended consequences in the future, ie. a too big radiation zone.

What I mean is this: if in the future some other GitHub API endpoint returns a non-RFC 2396 compliant URL, this workaround will hide and auto-correct the anomaly, by applying a fix that may or may not be appropriate for the issue at hand.

If the fix is appropriate, then so much the better, as no one will notice a thing. If however the fix is not appropriate for that future, for-now-hypothetical situation, then I fear it will be an arduous task to investigate and debug it.

I generally adopt a "Let It Rip!" approach: if that hypothetical situation arises at some point in the future, then let it rip - ie. let it throw an exception - so that we are made aware of it, can investigate it and can fix it in the most appropriate fashion.

Another way to look at it: in sporkmonger/addressable#205, the contract between Adressable and its clients was tightened to insist on RFC 3986 compliant schemes in all URIs, but the changes in this PR now relax that contract again, not only for the problematic `ssh_url` data point, but for **all** URIs processed by `octokit.rb`

You wrote:
> The approach in this PR simply handles the error and returns the raw value until we can investigate a better fix. Would this work for you?

I appreciate that my arguments are a tad philosophical and hypothetical at this point, so from a practical point of view, I'm happy to :shipit: (_but I am looking forward to the better fix you hint at_ :wink:)

pengwynn:
> AFAICT, the issue I discovered and reported in #727 does not point at a systemic and general issue with Sawyer::Relation#href, but rather at a very specific and localized issue with the ssh_url field returned by the various repository endpoints in the GitHub API and "consumed"/"processed" by octokit.rb.

@pvdb indeed! I think we'll roll with a more targeted approach in 5148fac.

pvdb:
Hey @pengwynn, I like it... the "radiation zone" has been drastically reduced in this latest version, which is awesome!

I guess the only comment I have - but just nitpicking, TBH - is that if and when other GitHub API endpoints _(other than the repository endpoints, that is)_ return or start returning a `ssh_url` data point, that too will be using the patched version of `Sawyer::Relation#href`, whether they need to or not.

ie. if that hypothetical API endpoint already returns a RFC 2396 compliant URL, then there isn't a need for using the patched version of the method; currently, however, it appears that [only the repository endpoints of the GitHub API return `ssh_url` fields](https://www.google.co.uk/search?q=%22ssh_url%22+site%3Adeveloper.github.com) so I'd say :shipit:

pvdb:
Quick afterthought... 

Instead of duplicating and modifying the implementation of `Sawyer::Relation#href` in `lib/ext/sawyer/relation.rb`, I wonder if it would be better to patch it outright, in order to future-proof it?

### pre-Ruby 2.0

```ruby
require 'sawyer'

module Sawyer
  class Relation
    super_href = instance_method(:href)
    define_method(:href) do
      # see: octokit/octokit.rb#727
      name.to_s == "ssh" ? @href : super_href.bind(self).()
    end
  end
end
```

### post-Ruby 2.0

```ruby
require 'sawyer'

module Patch
  def href(options=nil)
    # see: octokit/octokit.rb#727
    name.to_s == "ssh" ? @href : super
  end
end

Sawyer::Relation.send(:prepend, Patch)
```

pengwynn:
@pvdb Awesome! Applied your change in 3731782. Will 🚢 when this goes green.

pvdb:
Ah right, I failed to notice the "Supported Ruby Versions" section in the README document, which is why I offered two versions, but yeah: Ruby 2.x makes things a lot cleaner!

pvdb:
:thumbsup:

PR #836, opened 2016-11-22
Avoid polluting global namespace

yuku-t:
My app has a `::Patch` class. This PR fixes name collision problem.


pengwynn:
Oops, missed a wrapping namespace on the refactor. Thanks for the _patch_. :bowtie:

PR #837, opened 2016-11-23
Fix passing options to all_organizations

sschuberth:
Passing "since" did not work when calling "all_organizations" as it was
not passed to "paginate".


pengwynn:
Nice find. 🍰

PR #839, opened 2016-11-27
Removing unused rspec.subject and fixing typos on a variable name

esparta:
The specs on `Octokit::RateLimit` define a subject that is never used. Also did fix for typo on `resonse_headers`


tarebyte:
Thanks @esparta!

PR #844, opened 2016-12-06
Run tests by Ruby 2.3.3 and ruby-head

ojiry:
#### Purpose

Test with latest ruby version.
Ruby 2.4 will be released at an early date, so I want to add ruby-head.

#### Changes

Modify ruby version of .travis.yml.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ rvm:
   - 2.0
   - 2.1
   - 2.2
-  - 2.3.1
+  - 2.3.3
+  - ruby-head
 
 bundler_args: --without development
 
@@ -28,6 +29,7 @@ matrix:
   allow_failures:
     - rvm: rbx-2
     - rvm: jruby
+    - rvm: ruby-head
 
 notifications:
   emails: false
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -6,7 +6,8 @@ rvm:
   - 2.0
   - 2.1
   - 2.2
-  - 2.3.1
+  - 2.3.3
+  - ruby-head
```
koic: `ruby-head` is under development that is not stable. I think that add `ruby-head` to `allow_failures`.

cf. https://github.com/octokit/octokit.rb/blob/3887fd52134f56032b871304e56bd0f1753b0dc5/.travis.yml#L27-L30

ojiry:
Thanks @koic ! You’re right.

.travis.yml
```diff
@@ -28,6 +28,7 @@ matrix:
   allow_failures:
     - rvm: rbx-2
     - rvm: jruby
+    - rvm: ruby-head
```
koic: `ruby-head` is also required in the`rvm` section. `ruby-head` is not running as follows.

https://travis-ci.org/octokit/octokit.rb/builds/181598215

ojiry: Oops!
Let me sleep on it.

ojiry:
I updated rspec version because Travis CI does not work.
https://github.com/octokit/octokit.rb/pull/844/commits/93219e1f10f6363f2616e35a4c4eb2c5ba8a2574

.travis.yml
koic: This looks good.

PR #851, opened 2017-01-05
Fix application requests for non dotcom hosts

### LINKED ISSUE #831: Incorrectly builds urls for Github Enterprise API endpoints
The default Github Enterprise installation exposes its API at endpoints
prefixed with `/api/v3`. The current version of Octokit, however, has a bug
where when configuring the `api_endpoint`, as [recommended](https://github.com/octokit/octokit.rb#working-with-github-enterprise), results in requests to an
incorrect url.

**Expected behavior**

Configuring the Octokit client's `api_endpoint` value as

```
client = Octokit::Client.new(
  api_endpoint: "https://somedomain.com/api/v3",
  client_id: "123"
)
```

then calling `client.check_application_authorization("some-token")`

should result in a request to

```
https://somedomain.com/api/v3/applications/abc/tokens/some-token
```

**Actual behavior**

A request is made to 

```
https://somedomain.com/applications/abc/tokens/some-token
```

Notice the omitted `/api/v3`

**Steps to reproduce**

We've put together a [repository](https://github.com/codeclimate/octokit-enterprise-bug) with a simple demonstration of this bug. Run `rspec spec/test.rb` from the root of this repository

**Investigation**

The problem can be traced down to where the request url is built in [Faraday](https://github.com/lostisland/faraday/blob/master/lib/faraday/connection.rb#L406)

On these lines, `base` is a `URI` instance and the `+` operator is defined on `URI` as an alias to `merge` which does not exhibit the same behavior as concatenation and is why the `/api/v3` portion of the path gets lost.

**Proposed resolution**

At the moment, I'm not certain how to correct this issue from Octokit's point of view, but am happy to help.

pengwynn:
Fixes #831, this change switches to a relative path so that URLs are built correctly on Enterprise (and lab) hosts.

/cc @tarebyte @GordonDiggs

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
 gem 'jruby-openssl', :platforms => :jruby
-gem 'rake'
+gem 'rake', '> 11.0.1', '< 12.0'
 
 group :development do
   gem 'awesome_print', :require => 'ap'
--- a/lib/octokit/client/authorizations.rb
+++ b/lib/octokit/client/authorizations.rb
@@ -152,7 +152,7 @@ module Octokit
         secret = opts.delete(:client_secret) || client_secret
 
         as_app(key, secret) do |app_client|
-          app_client.get "/applications/#{client_id}/tokens/#{token}", opts
+          app_client.get "applications/#{client_id}/tokens/#{token}", opts
         end
       end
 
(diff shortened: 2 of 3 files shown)


pengwynn:
This checks out locally. Looks like we need to troubleshoot CI on Travis, @tarebyte.

tarebyte:
Nice @pengwynn! I totally thought that we were going to have to fix this at a higher level.

I'll take a look at the CI errors.

tarebyte:
:shipit:

PR #858, opened 2017-01-24
Don't mutate the options variable passed to API calls

jpmckinney:
Otherwise, you can't do this:

```ruby
headers = {accept: 'application/vnd.github.drax-preview+json'}
response = Octokit.repository_license_contents('octokit/octokit.rb', options)

 # `options` in this second call has been mutated to `{}`.
response = Octokit.repository_license_contents('octokit/octokit.py', options)
```


zenspider:
BUMP

tarebyte:
Looks fine to me, thanks @jpmckinney!

PR #860, opened 2017-01-27
Implement Review Preview API

soudy:
This PR implements the new reviews API https://developer.github.com/v3/pulls/reviews/.


soudy:
I think this is ready for review, could someone take a look?

tarebyte:
@soudy I should be able to look at this later today, thanks for the pull request!

lib/octokit/client/reviews.rb
```diff
@@ -0,0 +1,184 @@
...
+      # @param review_id [Integer] The id of the review
+      # @see https://developer.github.com/v3/pulls/reviews/#delete-a-pending-review
+      #
+      # @example
+      #   @client.delete_pull_request_review('octokit/octokit.rb', 825, 6505518)
+      #
+      # @return [Boolean] True if deleted, false otherwise
+      def delete_pull_request_review(repo, pull_id, review_id, options = {})
+        options = ensure_api_media_type(:reviews, options)
+        boolean_from_response(
```
tarebyte: Since the API returns a payload (https://developer.github.com/v3/pulls/reviews/#delete-a-pending-review) this shouldn't be a `boolean_from_response`. Those should be reserved for requests that respond with a `204`.

See comment in `/lib/octokit/connection.rb`

soudy:
@tarebyte Thanks for your feedback, I've updated the code.

soudy:
Anyone for a round 2? :smile:

tarebyte:
I can take another look, @soudy.

If I might ask a favor though, would you mind not force pushing onto the branch anymore for this PR?

For someone who needs to review the work it's hard to keep track of what I'm reading when the code keeps constantly changing on "the same commit".

Please and thank you ✨

soudy:
Ah sorry @tarebyte, will do. I do try to keep them organized but you're right about the changing.

lib/octokit/client/reviews.rb
```diff
@@ -0,0 +1,174 @@
...
+      # @return [Array<Sawyer::Resource>] Array of Hashes representing the reviews
+      def pull_request_reviews(repo, id, options = {})
+        options = ensure_api_media_type(:reviews, options)
+        get "#{Repository.path repo}/pulls/#{id}/reviews", options
+      end
+
+      # Get a single review
+      #
+      # @param repo [Integer, String, Hash, Repository] A GitHub repository
+      # @param pull_id [Integer] The id of the pull request
```
tarebyte: Can we update all of the uses of `pull_id` to use `number` instead. It makes it consistent with the rest of the API methods.

lib/octokit/client/reviews.rb
```diff
@@ -0,0 +1,174 @@
...
+      # @param review_id [Integer] The id of the review
```
tarebyte: Could we also update all uses of `review_id` to be `review`. This makes it more consistent with the rest of the library.

pulusanidamini:
Hi @soudy .. Any idea when this PR will be merged ?

soudy:
@pulusanidamini I just implemented @tarebyte's last feedback, should be good to go now I think. :smile:

tarebyte:
Thanks for working on this @soudy 💖

bbuchalter:
I was just trying this out but did not get the results I expected. See attached screenshots.
screen shot 2017-03-04 at 1 23 29 am
screen shot 2017-03-04 at 1 23 06 am

Please let me know if I can provide any other detail, or if you prefer a different  forum for this report. Thanks for your work on octokit!

bbuchalter:
Upon further investigation, the PR I was testing with was merged and the requested review had been completed. Perhaps some documentation about what states the requests must be in to be returned in the API would be helpful?

PR #864, opened 2017-02-09
Fix update_pull_request doc

chrisarcand:
The example method signature incorrectly uses 'id' instead of 'number', the name of the detailed parameter given. The PR number and not it's ID is the actual thing used, so that was confusing to read at first.


tarebyte:
🍰

Issue #832, opened 2016-11-13
Integrations endpoints


tommoor:
It would be great to see support added for the three new integrations endpoints that are currently in `early release` status…

https://developer.github.com/v3/integrations

PR #871, opened 2017-03-06
Add support to prerelease Integrations API

tarebyte:
This adds the early access Integrations API (https://developer.github.com/v3/integrations/) to Octokit.

Closes https://github.com/octokit/octokit.rb/issues/832

/cc @keavy for Integration 👀


lib/octokit/client/integrations.rb
```diff
@@ -0,0 +1,77 @@
...
+      # Create a new installation token
+      #
+      # @param installation [Integer] The id of a a GitHub Integration Installation
+      # @param options [Hash] An customizable set of options
+      #
+      # @see https://developer.github.com/v3/integrations/#find-installations
+      #
+      # @return [<Sawyer::Resource>] An installation token
+      def create_integration_installation_access_token(installation, options = {})
+        options = ensure_api_media_type(:integrations, options)
```
joeyw: As a pattern across the library, we avoid mutating the parameters. In other methods we usually use `opts` as the variable for options after mutating them. In this case it would unlikely cause issues, but would bring everything into parity to follow the same pattern.

spec/octokit/client/integrations_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+          end
+        end # .remove_repository_from_integration_installation
+      end # with installed repository on installation
+    end # with repository
+  end # with integration installation
+
+  private
+
+  def new_jwt_token
+    private_pem = File.read(ENV['OCTOKIT_TEST_INTEGRATION_PEM_KEY'])
```
joeyw: This is throwing a nil error when the env variable is not defined. This could be changed to follow the rest of the library, using fetch inside a helper method in the spec helper. The fixture used for the CI could be used for it.

joeyw:
@tarebyte Aside from my nitpicking, this looks good after a quick one-over. You seem to have a firm grasp of the vcr placeholders too and how to exploit them gainfully, 👍

lib/octokit/client/integrations.rb
tarebyte: Done in e0d5884

spec/octokit/client/integrations_spec.rb
tarebyte: Done in 7a5940c

ceslami:
Thanks @tarebyte!

PR #875, opened 2017-03-10
Feature to retrieve the list of required status for a protected branch

kevinrobayna:
Edit: I fix my previous problem but now I have another failure because in theory the end point to list_required_status requires something in the headers that is not listed in the website they are pointing to.

Anyone can help me?

```
     Failure/Error: @client.list_required_status(test_repo, branch)
     Octokit::UnsupportedMediaType:
       GET https://api.github.com/repos/kevinrobayna/octokit.rb/branches/master/protection/required_status_checks/contexts: 415 - If you would like to help us test the Protected Branches API during its preview period, you must specify a custom media type in the 'Accept' header. Please see the docs for full details. // See: https://developer.github.com/v3/
```


kevinrobayna:
@jcastillocano @agarfu

kytrinyx:
This is failing because it is missing the preview_header as part of the method definition.

See this example:
https://github.com/octokit/octokit.rb/blob/master/lib/octokit/client/apps.rb#L15

kytrinyx:
@kevinrobayna Since we've not heard back since July, I'm going to go ahead and close this. If you do want to move this forward, please open a new pull request. 🌷

PR #876, opened 2017-03-23
Remove destructive assignment for repository

pocke:
`Octokit::Repository` breaks an argument if the argument is a Hash.
This change fixes the problem.

# Reproduce

```ruby
require 'octokit'

repo = {name: 'foo', username: 'bar'}

Octokit::Repository.new(repo)

p repo # => {:name=>"foo", :username=>"bar", :repo=>"foo", :user=>"bar", :owner=>"bar"}
```


pocke:
ping

kytrinyx:
I think this is a very reasonable change.

Some people may be relying on the mutation, which I would consider a bug. My suggestion for this would be to do a patch release with this change.

PR #877, opened 2017-03-23
Make to understandable error message when `Octokit::Repository.new`is failed

pocke:
Note: Test cases for this PR is failed now. It will be fixed if #876 was merged.

Currently, an error message of `Octokit::Repository.new` is unkind.

For example:

```ruby
Octokit::Repository.new(nil) # => Octokit::InvalidRepository: Invalid Repository. Use user/repo format.
```

If an user want to pass an id(Integer), this message is wrong for the user. Because the user doesn't expect the `user/repo` format.

This change improves the error message.
So, this change makes debugging easier.

### DIFF
--- a/lib/octokit/repository.rb
+++ b/lib/octokit/repository.rb
@@ -31,10 +31,10 @@ module Octokit
         @name = repo[:repo] || repo[:name]
         @owner = repo[:owner] || repo[:user] || repo[:username]
       else
-        raise_invalid_repository!
+        raise_invalid_repository!(repo)
       end
       if @owner && @name
-        validate_owner_and_name!
+        validate_owner_and_name!(repo)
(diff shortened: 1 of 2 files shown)


kytrinyx:
@pocke this is a good observation, thanks!

I think this is the right approach, however I'd like to see if we can find a more fluid way of phrasing the error message.

How about:

> #{repo.inspect} is invalid as a repository identifier. Use the repo/user (string) format, or the repository ID (integer), or a hash containing `:repo` and `:user` keys.

pocke:
@kytrinyx Thanks! Seems good to me. I replaced the error message with it. 
And I rebased this branch.

kytrinyx:
This looks great, thank you.

PR #878, opened 2017-03-28
perform a merge

flyerhzm:
see https://developer.github.com/v3/repos/merging/#perform-a-merge


pengwynn:
Thanks for the patch, but this is already implemented [here](https://github.com/octokit/octokit.rb/blob/c1e66093eeb499c59e49212ca10c2a4efc356836/lib/octokit/client/commits.rb#L201-L215).

PR #879, opened 2017-03-29
Fix Gundam in the README

orta:
Perhaps an automation went wrong? In any case, this brings it back.


pengwynn:
Hmm, wonder if it's related to [this](https://github.com/blog/2333-a-formal-spec-for-github-flavored-markdown). Thanks for the patch. 🍰

PR #880, opened 2017-03-30
Fix incorrect mimetype for integrations in PREVIEW_TYPES

pior:
Just a small fix related to the new *Integrations*.


tarebyte:
:cake:

PR #881, opened 2017-04-02
Change several occurrences of `Hash#merge!`

### LINKED ISSUE #813: Stop mutating parameters
Change all methods similar to #780 which dups parameter before mutation. There are many more instances in octokit that mutate parameters we should change to use this approach.

davidcpell:
Fixes #813

I had originally opened PR #780, which is mentioned in #813, and
thought I would take a look through the repo for other places where mutating
opts args could be avoided.


pengwynn:
🍰

PR #882, opened 2017-04-11
Do not add an empty commit_message in the merge_pull_request api call

AgarFu:
If you want to squash a PR the default commit message set by GitHub is no being set because octokit always put a custom commit_message.

This patch removes the commit_message from the body if the commit message is nil.


kytrinyx:
This is a breaking change, however I do think that it's worth making it when we are ready to do our next major release.

The change that we could make for now is to add a deprecation notice if someone is calling the endpoint without a message. See https://github.com/octokit/octokit.rb/blob/c3aa3e057f629a8176b4ea9b8615b25352604c5a/lib/octokit/client/apps.rb for an example of how we implement deprecation notices.

kytrinyx:
@AgarFu since we've not heard back since July, I'm going to go ahead and close this.

If you do want to move this forward, please submit a new PR with the deprecation notice.

PR #883, opened 2017-04-15
Handle unauthorized exception when checking repo

gylaz:
If a token does not have access to a repo (say it was deleted on
GitHub's side), accessing a repo will throw a `401` (Bad credentials).

Not totally sure if it makes sense for the `repository?` call, since
technically the repository exists but the token isn't priviledged.


mommahost:
P

On Apr 14, 2017 11:19 PM, Coveralls <notifications@github.com> wrote:

[Coverage Status]<https://coveralls.io/builds/11096042>

Coverage increased (+0.0007%) to 99.26% when pulling 1208a7b<https://github.com/octokit/octokit.rb/commit/1208a7b7a1386aaf4bda1cc954565a0fbd8fe5f7> on gylaz:gl-rescue-unauthorized into 7bc49be<https://github.com/octokit/octokit.rb/commit/7bc49beb839818974e6c8aa568a348bd5f2cca54> on octokit:master.

kytrinyx:
If the token doesn't exist, then 401 Bad Credentials is the correct response.

I'm going to go ahead and close this, but if I've misunderstood, please let us know so we can discuss.

PR #885, opened 2017-04-22
Add 'topics' (preview)

koenrh:
In January, GitHub introduced [topics](https://github.com/blog/2309-introducing-topics) (labels for repositories). This property is returned in preview-mode (if the accepted media type includes `application/vnd.github.mercy-preview+json`) for repository objects.

Even though topics are already included in responses if you opt-in to this preview feature, this adds the media type to the `PREVIEW_TYPES` hash so that users will get the preview warning.

I'm not sure if the added spec is even needed.


pengwynn:
🍰

PR #886, opened 2017-04-26
Allows reviews and review comments to be paginated

alindeman:
Per <https://developer.github.com/v3/pulls/reviews/> and <https://developer.github.com/v3/pulls/review_requests/>, these routes can be paginated.


pengwynn:
Thanks @alindeman. 🍰

PR #892, opened 2017-05-23
Move integrations endpoints across to apps

greysteil:
GitHub Integrations are dead. Long live GitHub Apps.

Renames client library module, and updates all endpoints to use apps, rather that integrations. Adds aliases for the original methods, so backwards compatibility isn't broken.


greysteil:
There are a couple of additional app methods I'd like to add (getting `/user/installations` is the most obvious current omission) but figured I'd get this in first. @tarebyte - any chance of a quick review on it?

lib/octokit/client.rb
```diff
@@ -22,7 +22,7 @@
 require 'octokit/client/gists'
 require 'octokit/client/gitignore'
 require 'octokit/client/hooks'
-require 'octokit/client/integrations'
+require 'octokit/client/apps'
```
tarebyte: Can you move this so that the list is in Alphabetical order again?

spec/octokit/client/apps_spec.rb
```diff
@@ -16,7 +16,7 @@
     it "returns installations for an integration" do
       installations = @jwt_client.find_integration_installations
       expect(installations).to be_kind_of Array
-      assert_requested :get, github_url("/integration/installations")
```
tarebyte: I'd really like to keep the old tests around so that when someone removes the aliases the tests notice.

I'm not too worried about duplication or anything like that.

greysteil: Done - I've added specs for the aliased methods, so we'll know if anyone removes them. (It doesn't affect this line, because both methods will now hit `/app/installations`, which I think is the right behaviour.)

greysteil:
Awesome - updated. Anyone else who could review while you're away?

greysteil:
@tarebyte - hope you had a great vacation! This should be good to go.

tarebyte:
I did thanks @greysteil! I had one more thought this morning.

Instead of just aliasing the methods I'd like to add a deprecation warning take a look at https://github.com/octokit/octokit.rb/blob/master/lib/octokit/client/organizations.rb#L308 as an example.

greysteil:
On it.

lib/octokit/client/apps.rb
```diff
@@ -0,0 +1,114 @@
...
+      #
+      # @return [Array<Sawyer::Resource>] A list of installations
+      def find_app_installations(options = {})
+        opts = ensure_api_media_type(:integrations, options)
+        paginate "/app/installations", opts
+      end
+      alias find_installations find_app_installations
+
+      def find_integration_installations(options = {})
+        octokit_warn "Deprecated: The find_integration_installations method is deprecated. Use find_app_installations instead."
```
tarebyte: Could we update the wording just a titch?

```
Octokit::Client::Apps#find_integration_installations method is deprecated. Please update your call to use Octokit::Client::Apps#find_app_installations before the next major Octokit version update.
```

greysteil:
Done!

greysteil:
🎉

PR #894, opened 2017-05-28
Add `Client.add_assignees`

tomb0y:
This pull request adds a new method that makes it possible to assign
users to issues.


tarebyte:
:cake:

PR #895, opened 2017-05-30
Add Octokit::Client::Apps#find_user_installations method

greysteil:
Specs will fail as I haven't added a cassette. Can someone who already has the required environment variable set up do so?


greysteil:
@tarebyte - any chance you could do me the honour of a cassette?

greysteil:
Now with passing cassette (I cheated, and borrowed the one for `find_app_installations` which has an identical response).

Bumped the Ruby 2.4 version on travis as I was getting segfaults on 2.4.0.

greysteil:
🎉

PR #898, opened 2017-06-07
Incorrect anchor link to add-create-membership

brntbeer:
the documentation previously linked to an anchor that didn't exist.

cc @gjtorikian or @tarebyte because i don't know who else to include because there's no maintainers list 😨 

❤️


tarebyte:
💖

PR #899, opened 2017-06-08
Expose response_headers and response_body on error objects

greysteil:
Previously, the only way to get at these was with `error.instance_variable_get(:@response)`, which is grim.


pengwynn:
🍰

PR #904, opened 2017-06-21
Add app user endpoints

seejohntest:
I debated putting these into a new `client/app_user`, but wanted to see what others thought first.

Will add VCR once we're settled, since it requires a lot of setup as far as I can see.


tarebyte:
@seejohnrun `find_user_installation` exists, but `find_installation_repositories_for_user` is not, mind merging master and making these changes?

kytrinyx:
Hey @seejohnrun I'm going to go ahead and close this one. If you have time to get back to it that would be ✨ and we can have another look at it.

Issue #906, opened 2017-07-13
Where should I add Marketplace API


sotayamashita:
I am implementing [marketplace API](https://developer.github.com/v3/apps/marketplace/).  Here is the URL: https://github.com/sotayamashita/octokit.rb/tree/feature/marketplace.

tarebyte:
@sotayamashita the best place would probably be in [this directory](https://github.com/octokit/octokit.rb/tree/ad74f84e818ee15d2e49422d51978a40b2d957a0/lib/octokit/client) under `marketplace.rb`

sotayamashita:
@tarebyte Thanks

PR #908, opened 2017-07-17
WIP: Add Octokit::Client::Marketplace#find_app_marketplace_purchases method

sotayamashita:
##  TODO

- [ ] Write test for it 

- https://developer.github.com/v3/apps/marketplace/#get-a-users-marketplace-purchases
- #906

### DIFF
--- a/lib/octokit/client.rb
+++ b/lib/octokit/client.rb
@@ -29,6 +29,7 @@ require 'octokit/client/legacy_search'
 require 'octokit/client/licenses'
 require 'octokit/client/meta'
 require 'octokit/client/markdown'
+require 'octokit/client/marketplace'
 require 'octokit/client/milestones'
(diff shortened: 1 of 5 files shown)


sotayamashita:
I am confusing because this Marketplace API says 

> Endpoints for `/marketplace_listing/*` and `/user/marketplace_purchases` are only accessible by OAuth Apps using `client_id` and `secret` basic authentication or ...

It means we need specify client_id and client_secret of OAuth app.

For example:

```
curl -u "<client_id>:<client_secret>" -H "Accept: application/vnd.github.valkyrie-preview+json" -H "Authorization: token <Personal access token>" "https://api.github.com/user/marketplace_purchases"
```

However, I can request the API without client_id and client_secret. Does I misunderstand?

PR #922, opened 2017-08-24
allow to use underscore in user and repository name

contribu:
github allow to use underscore in user and repository name.
However octokit doesn't.
I fixed to allow to use underscore in user and repository name.


contribu:
I was confused.

PR #926, opened 2017-09-02
Delete docs for defunct options for marking thread as read

nodunayo:
As per the [https://developer.github.com/v3/activity/notifications/#mark-a-thread-as-read](https://developer.github.com/v3/activity/notifications/#mark-a-thread-as-read), this endpoint no longer recognises any options passed to it.

@andrew and I attempted to change a notification's unread status, via the optional argument, but this was not possible.


andrew:
The CI failure seems unrelated to this change

tarebyte:
💖

PR #928, opened 2017-09-06
Add support for the new Nested Teams API preview

LizzHale:
This PR adds support for  [the Nested Teams API preview](https://developer.github.com/changes/2017-08-30-preview-nested-teams/):

* Adds the new `child_teams` method for listing a team's child teams
* Updates the documentation for `create_team` and `update_team` to include the `parent_team_id` option.


lib/octokit/client/organizations.rb
```diff
@@ -292,6 +292,7 @@ def organization_teams(org, options = {})
       # @option options [String] :name Team name.
       # @option options [Array<String>] :repo_names Repositories for the team.
       # @option options [Array<String>] :maintainers Maintainers for the team.
+      # @option options [Integer] :parent_team_id ID of a team to set as the parent team.
```
tarebyte: What do you think about either adding the preview to these methods, or mentioning in the comments that preview is required for this option?

LizzHale: Great point @tarebyte. I updated the PR to ensure the `:nested_teams` media type when the `parent_team_id` option is passed in: a9793a0. Let me know if that's what you had in mind.

PR #932, opened 2017-09-14
Add new apps endpoint

seejohnrun:
Adds `find_installation_repositories_for_user`

Replacement for https://github.com/octokit/octokit.rb/pull/904

<strike>Will add VCR once we're settled.</strike>


spec/octokit/client/apps_spec.rb
```diff
@@ -49,6 +49,14 @@
       end
     end # .installation
 
+    describe ".find_installation_repositories_for_user", :vcr do
```
kytrinyx: You shouldn't need to pass the `:vcr` option here.

seejohnrun:
what do you think @kytrinyx ?

lib/octokit/client/apps.rb
```diff
@@ -146,6 +146,19 @@ def remove_repository_from_integration_installation(installation, repo, options
         )
         remove_repository_from_app_installation(installation, repo, options)
       end
+
+      # List repositories accessible to the user for an installation
+      #
+      # @param installation [Integer] The id of a a GitHub App Installation
```
kytrinyx: "a" is here twice.

seejohnrun:
@kytrinyx Fixed that another another grammatical error that were present in multiple places in this file - thanks!

PR #948, opened 2017-10-16
Update docs to indicate Ruby < 2.2 not supported

toddmowen:
#924 dropped testing against Ruby 2.0 and 2.1, but the README still claimed that these versions were supported. In fact, a project I was building on Travis with Ruby 2.0.0 stopped working:

```
$ gem install octokit netrc
YAML safe loading is not available. Please upgrade psych to a version that supports safe loading (>= 2.0).
Fetching: multipart-post-2.0.0.gem (100%)
Successfully installed multipart-post-2.0.0
Fetching: faraday-0.13.1.gem (100%)
Successfully installed faraday-0.13.1
Fetching: public_suffix-3.0.0.gem (100%)
ERROR:  Error installing octokit:
	public_suffix requires Ruby version >= 2.1.
```

Question: Should `required_ruby_version` in `octokit.gemspec` be bumped too? It's currently set to 2.0.0.

/cc @tarebyte

### DIFF
--- a/README.md
+++ b/README.md
@@ -653,8 +653,6 @@ when writing new specs.
 This library aims to support and is [tested against][travis] the following Ruby
 implementations:
 
-* Ruby 2.0
-* Ruby 2.1
 * Ruby 2.2
 * Ruby 2.3
 * Ruby 2.4


kytrinyx:
> Question: Should required_ruby_version in octokit.gemspec be bumped too? It's currently set to 2.0.0.

That's a good question. If we've dropped support in the tests, I would suggest that we bump it everywhere.

toddmowen:
Further delving into the reason for my project failing with Ruby 2.0, it seems like this was triggered by the release of [addressable 2.5.2](https://rubygems.org/gems/addressable/versions/2.5.2) on August 25, which declared compatibility with the 3.x releases of [public_suffix](https://rubygems.org/gems/public_suffix) (public_suffix 3.0.0 had just been released a few weeks earlier). Unfortunately, public_suffix 3.0.0 raised the required Ruby version from 2.0 to 2.1.

I'm going to report this to the author of addressable, since it is technically a violation of their semantic versioning policy.

toddmowen:
I agree that octokit.rb should bump the required Ruby version to >= 2.2. However, for the same reason that I argue that addressable has violated semantic versioning, doing so would be a backwards-incompatible change and thus require bumping the version to 5.0.0. I'm not in a position to make such a change.

kytrinyx:
> I agree that octokit.rb should bump the required Ruby version to >= 2.2. However [...] doing so would be a backwards-incompatible change and thus require bumping the version to 5.0.0

Yes, I agree. We're going to need to hold off on doing that until we do a major release.

toddmowen:
@dentarg persuaded me that the root cause of my original build failure is actually rubygems, since it attempts to install a version of public_suffix that is incompatible with the version of Ruby I was running. This is a known issue: rubygems/rubygems#968

So the current octokit.gemspec is _technically_ correct in advertising compatibility with Ruby >= 2.0.

I still suggest merging the current PR, so that the docs indicate that Ruby 2.0 and 2.1 are not officially supported (since we no longer include them in the test matrix).

tarebyte:
@toddmowen for the sake of keeping everything compatible we will add back in testing for Ruby 2.0 and 2.1 for the 4.8.0 release. When we cut for 5.0.0 we will go back and remove support for 2.0 and 2.1 again.

I'm going to leave this open for now and merge this in when we're ready to ship 5.0.0.

timrogers:
We now officially only support Ruby 2.7 onwards, so this PR can be closed. Thanks for your work here to keep our documentation honest ❤️

PR #955, opened 2017-10-25
Revert "fixes contents method example on README [ci skip]"

beodeutk:
Reverts octokit/octokit.rb#953


tarebyte:
@beodeutk can you give us more context as to why you want to revert?

PR #959, opened 2017-11-02
Add support for the Commit Search API preview

azizshamim:
This PR adds support for the [Commit Search API preview](https://developer.github.com/v3/search/#search-commits):

* Adds a new `search_commits` method for returning the search results for commits
* Adds documentation for the new method (it's basically the same as the other search methods)


azizshamim:
I know I need a new cassette. The code works in `script/console`. Not entirely sure why a new cassette isn't being generated when I run the tests.

azizshamim:
@tarebyte - any tips on getting the right fixtures into this branch? I'm struggling generating them.

tarebyte:
@azizshamim everything seems to be ok, I'll take a look to see if something funky is going on.

azizshamim:
@tarebyte I'm about 90% sure it's because I don't have the right credentials set up, nor the right configuration set up, to connect to GitHub API and generate the right fixture for the tests. Any guidance would be much appreciated.

lib/octokit/client/search.rb
```diff
@@ -20,6 +20,21 @@ def search_code(query, options = {})
         search "search/code", query, options
       end
 
+      # Search commits
+      #
+      # @param query [String] Search terms and qualifiers
+      # @param options [Hash] Sort and pagination options
+      # @option options [String] :sort Sort field
```
cheshire137: Maybe `Field to sort by` instead of `Sort field`.

azizshamim: I was matching the language of the other option descriptions.

kytrinyx: Yeah, the rest of the codebase uses `Sort field`. Some endpoints also mention what the available sort options are, but many do not.

kytrinyx:
@azizshamim Did you define `OCTOKIT_TEST_GITHUB_TOKEN` with a valid personal token when you ran the tests?

azizshamim:
@kytrinyx - Yes. Just retested it, same error.

```ruby
  1) Octokit::Client::Search.search_commits searches commits
     Failure/Error: results = @client.search_commits 'repo:octokit/octokit.rb author:jasonrudolph', \
     NoMethodError:
       undefined method `<<' for {:read_timeout=>60, :continue_timeout=>nil, :debug_output=>nil}:Hash
       Did you mean?  <
```

Does the token need any specific permissions?

kytrinyx:
> Does the token need any specific permissions?

No, it shouldn't. Let me pull this down and see if I can figure out what's going on.

kytrinyx:
This worked for me with a token that didn't have any scopes. I've pushed up the VCR cassette. If this passes I'll go ahead and merge it.

kytrinyx:
Thanks for your patience in getting this change in! ✨

azizshamim:
Thanks for the assist @kytrinyx. I wonder if I’ve borked my global `vcr` setup somehow.

I’d like to add support for all the API previews that aren’t implemented to this gem. I wonder if I can fix it before I open the next Pull Request.

kytrinyx:
One thing occurred to me... did you delete the generated cassette between
each attempt? One thing that I tend to do is retry... after a failed call
gets recorded by VCR, thereby replaying the bad request /facepalm

azizshamim:
> did you delete the generated cassette between each attempt?

The cassette never regenerates if I delete it. None of them do. Whether I have a valid token in `OCTOKIT_TEST_GITHUB_TOKEN` or not.

kytrinyx:
Ok, that totally sucks! I have no idea what might be going on here.

tarebyte:

Issue #952, opened 2017-10-23
Release v4.8.0?


greysteil:
It's been a while since the last release (which was in March) and there are a *lot* of [changes on master](https://github.com/octokit/octokit.rb/compare/v4.7.0...master). Would be rad to get them released - is there anything blocking?

tarebyte:
We agree that we need to cut a release, sorry it's been a while 😅 

# TODO
- [ ] Write up the changelog
- [ ] Ensure there aren't any breaking changes
- [ ] If there breaking changes cut a branch release candidate for 4.8.0 and another for 5.0.0
- [ ] Run rake task to send release(s) to Rubygems and GitHub
- [ ] Post change log to release notes

kytrinyx:
## TODO
https://github.com/octokit/octokit.rb/pull/892 removes support for the deprecated Integrations namespace, replacing it with `Apps`. I think we need to assign `Apps` to a new constant `Integrations` so that we don't cause a backwards-incompatible change.

## Changelog notes

Using https://github.com/octokit/octokit.rb/compare/v4.7.0...master

We have been unable to get a rubinius build working for CI for quite some time, meaning that rubinius support has been spotty or non-existent. We've stopped trying. Not backwards-incompatible directly since it already wasn't working, but a little bit iffy anyway.

**Add support for additional API endpoints**
* Add support for Topics #885
* Add support for Get a single Installation #884
* Add support for Add assignees #894
* Add support for Find user installations #895
* Add support for marketplace listings API #929
* Add support for the new Nested Teams API preview #928
* Add support for Get user's marketplace purchases #938
* Add support for new apps endpoint #932

**Tweak behavior**
* Allow reviews and review comments to be paginated #886
* Expose response_headers and response_body on error objects #899
* Remove destructive assignment for repository #876
* Improve error message when `Octokit::Repository.new` fails #877
* Take the Organization Memberships API out of preview #913
* Take the Pull Request Review API out of preview #923
* Update add/remove installations methods to use non-deprecated endpoints #958
* Fix apps client for GitHub Enterprise installs (new endpoints in this release, but still) #966

kytrinyx:
Wait. It says that #892 aliases the namespace. I can't see which line. Looking more carefully.

kytrinyx:
Oh, I see what's going on. We don't actually reference the namespace directly (e.g. Octokit::Client::App). So this is fine.

Sorry about the kerfuffle!

kytrinyx:
[...]

PR #962, opened 2017-11-17
Add Ruby 2.0 and 2.1 back into the Travis matrix

tarebyte:
This reverts https://github.com/octokit/octokit.rb/pull/924 so that we have compatibility for the upcoming [4.8.0 release](https://github.com/octokit/octokit.rb/issues/952#issuecomment-344950058).

When we ship 5.0 we are going to officially remove support for these old versions 😄

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ language: ruby
 rvm:
   - jruby
   - rbx-2
+  - 2.0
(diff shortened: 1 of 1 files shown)


tarebyte:
@kytrinyx as soon as the builds pass, feel free to 🚢

PR #966, opened 2017-12-14
Fix apps client for GitHub Enterprise installs

jdennes:
This shows and fixes the fact that the client doesn't work correctly for the app endpoints documented at https://developer.github.com/v3/apps/ when used in a GitHub Enterprise environment.

Without this change requests like `POST https://ghe.localhost/installations/:installation_id/access_tokens` are being made instead of `POST https://ghe.localhost/api/v3/installations/:installation_id/access_tokens`, for example.

I had to manually patch the client to work with Octokit.rb using these endpoints on a GitHub Enterprise install.

Seems similar to https://github.com/octokit/octokit.rb/issues/831 + https://github.com/octokit/octokit.rb/pull/851.

### DIFF
--- a/lib/octokit/client/apps.rb
+++ b/lib/octokit/client/apps.rb
@@ -13,7 +13,7 @@ module Octokit
       # @return [Array<Sawyer::Resource>] A list of installations
       def find_app_installations(options = {})
         opts = ensure_api_media_type(:integrations, options)
-        paginate "/app/installations", opts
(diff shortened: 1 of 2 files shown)


kytrinyx:
Thank you so much for figuring this one out. @tarebyte and I spent some time a couple of weeks ago trying to chase something similar, and we ended up waaaay out in the weeds. 🌿

PR #969, opened 2017-12-20
Document the steps necessary to cut a release

kytrinyx:
This is nothing new or special, but it should provide a handy checklist for cutting new releases,
which in turn might cause us to do this more often.

The gap between the v4.7.0 and v4.8.0 releases was about 8 months, mostly because we had some
turnover in maintainers, and the new maintainers had never cut a release, so it loomed as a somewhat uncomfortable unknown.

Documenting this should make it easier for new maintainers to whip up a release when it makes sense to do so.

[View rendered checklist](https://github.com/octokit/octokit.rb/blob/kytrinyx/release-docs/RELEASE.md)

@pengwynn since you've probably done most of the releases until now, would you have 5 minutes to glance through this checklist and make sure I'm not missing anything?

/cc @bkeepers as unofficial cat-herder of the octokit libraries

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,28 @@
+# Releasing a new version of octokit.rb
+
+1. Create a list of all the changes since the prior release
+    1. Compare the latest release to master using https://github.com/octokit/octokit.rb/compare/`${latest}`...master
+    1. Open the linked pull requests from all the `Merge pull request #...` commits
+    1. For all non-documentation PRs, copy title (including pull request number) into markdown list items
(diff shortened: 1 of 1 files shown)


pengwynn:
Yep, that looks right. 👍  I _think_ @tarebyte is set up to add new maintainers, but let me know if you need me to add anyone to Rubygems.

kytrinyx:
@pengwynn Yepp, @tarebyte added me yesterday so we're all set.

tarebyte:

PR #972, opened 2018-01-02
Url to the review API was wrong

Edouard-chin:
👋  Stumble across a small url mistake on the pull review api url

Thanks for this great gem!


tarebyte:
Thanks @Edouard-chin!

tarebyte:

PR #973, opened 2018-01-05
Fix pull_requests_comments example documentation

robd:
The `sort` and `direction` options had incorrect values. Update them to be consistent with the intent of the example.


tarebyte:
Thanks @robd!

tarebyte:

PR #976, opened 2018-01-15
Fix pagination of GitHub App endpoints

### LINKED ISSUE #974: Pagination fails for some GitHub App APIs
It seems that Octokit is unable handle pagination for some of the GitHub Apps API. It only returns the result for the first page. Even if there are more items.

Take the APIs to list repositories as an example:

Pagination expects `last_response.data` to be of Type Array, but App API returns Sawyer::Resource instead. Therefore concat of subsequent responses fails:

```
# Octokit::Connection
def paginate
#...
  data.concat(@last_response.data) if @last_response.data.is_a?(Array)
```
https://github.com/octokit/octokit.rb/blob/master/lib/octokit/connection.rb#L92

When comparing the API responses for GitHub App and Oauth App, it gets obvious:

1. List repositories of an installation (GitHub App) https://developer.github.com/v3/apps/installations/#list-repositories
```
{
  "total_count": 1,
  "repositories": [
    {
      "id": 1296269,
      ...
    }
  ]
}

=> Hash {}
```

2. List organization repositories (OAuth App) https://developer.github.com/v3/repos/#list-organization-repositories
```
[
  {
    "id": 1296269,
    ....
    }
  }
]

=> Array []
```

stmllr:
Fixes #974


stmllr:
I have no clue why the ruby 2.4.3 job failed at travis https://travis-ci.org/octokit/octokit.rb/jobs/329094824 Works for me, tested on the ruby:2.4.3-alpine3.7 -->  2.4.3p205 (2017-12-14 revision 61247) [x86_64-linux-musl]

timhaines:
Thanks for the PR @stmllr! Pinging @tarebyte.  Would love to see it merged and shipped.  We've had a few customers run into issues due to this bug.

stmllr:
ping

lib/octokit/client/apps.rb
```diff
@@ -33,10 +33,12 @@ def find_integration_installations(options = {})
       #
       # @see https://developer.github.com/v3/apps/#list-installations-for-user
       #
-      # @return [Array<Sawyer::Resource>] A list of installations
+      # @return [Sawyer::Resource] A list of installations
```
kytrinyx: I believe this returns a `Sawyer::Resource` containing a `total_count` and installations`, which is an array.

lib/octokit/client/apps.rb
```diff
@@ -80,10 +82,12 @@ def create_integration_installation_access_token(installation, options = {})
       # @param options [Hash] A customizable set of options
       # @see https://developer.github.com/v3/apps/installations/#list-repositories
       #
-      # @return [Array<Sawyer::Resource>] A list of repositories
+      # @return [Sawyer::Resource] A list of repositories
```
kytrinyx: Likewise here, I think this actually returns a resource containing a `total_count` and a list of `repositories`.

lib/octokit/client/apps.rb
```diff
@@ -154,10 +158,12 @@ def remove_repository_from_integration_installation(installation, repo, options
       #
       # @see https://developer.github.com/apps/building-integrations/setting-up-and-registering-github-apps/identifying-users-for-github-apps/
       #
-      # @return [Array<Sawyer::Resource>] A list of repositories
+      # @return [Sawyer::Resource] A list of repositories
```
kytrinyx: Here, too.

I realize that these were already wrong, and you made them less wrong—but since this is being tweaked, I'd love to get them right.

spec/octokit/client/apps_spec.rb
```diff
@@ -60,6 +60,15 @@
 
       assert_requested request
     end
+
+    it "utilizes auto_pagination", :vcr do
```
kytrinyx: I'm wondering if `allows` would be clearer than `utilizes`. I'd be afraid that the endpoint automatically is auto-paginated, whereas this requires the user to set `auto_paginate` to true.

spec/octokit/client/apps_spec.rb
```diff
@@ -60,6 +60,15 @@
...
+      @client.auto_paginate = true
+      response = @client.find_user_installations(accept: preview_header, per_page: 1)
+
+      expect(response.total_count).to eq 2
+      expect(response.installations.count).to eq 2
+      expect(response.installations).to be_kind_of(Array)
```
kytrinyx: The test itself looks good to me ⚡️

PR #980, opened 2018-01-22
Change the default value of `update_ref` `force` to false

pocke:
Problem
====

Currently, `Octokit::Client#update_ref` updates a reference with `force: true` option. It works like `git push --force`.
I think the behaviour is a problem for two reasons.

Firstly, the behaviour is dangerous.
Force push destroys pushed commits sometimes. I think dangerous operations should be optional.

Secondly, the default value is false in the GitHub API.

> Indicates whether to force the update or to make sure the update is a fast-forward update. Leaving this out or setting it to false will make sure you're not overwriting work. Default: false
>
> https://developer.github.com/v3/git/refs/#update-a-reference

I think Octokit default value should be same as the GitHub API.

Note
===

It is a breaking change. If someone uses `update_ref` method without a default value and expects "force push", it will be broken.
But I think we should change default behaviour. What do you think?


tarebyte:
Thanks for the PR @pocke!

I like this change, but like you said it's a breaking change to the client. So we're going to wait to 🚢 this when we start the process of releasing Octokit 5.0.

Let me know if you have any questions.

pocke:
> I like this change, but like you said it's a breaking change to the client. So we're going to wait to 🚢 this when we start the process of releasing Octokit 5.0.


It seems good. Thank you!

butly1721:
890

kfcampbell (approved):
I love the idea of matching the GitHub API's defaults, and we recently merged in another breaking change in #1495, so I think this is an opportune time to get this change in.

Issue #981, opened 2018-01-24
Can't set SSL verify mode in faraday options


taquitos:
When people use proxies for connection, they might want to disable SSL verification. This often helps during debugging when using a proxy like [Charles Proxy](https://www.charlesproxy.com).

The primary driver for this feature is to enable developers to use a proxy while they are debugging their own application that makes use of octokit.

taquitos:
Here's an example of usage:
```
if ENV['PROXY']
  Octokit.configure do |c|
    # for debugging only
    # This enables tracking of networking requests using Charles Web Proxy
    c.proxy = "https://127.0.0.1:8888"
    c.ssl_verify_mode = OpenSSL::SSL::VERIFY_NONE
  end        
end
```

You can see how we do something very similar in _fastlane_ [client.rb usage of verify mode](https://github.com/fastlane/fastlane/blob/1a79353a6d918eddca61274e97962a5f2770210e/spaceship/lib/spaceship/client.rb#L220-L224)

BasThomas:
Think this can be closed as #982 has been merged. :)

taquitos:
Sorry about that, thank you! 🎈:goat:

PR #982, opened 2018-01-24
Be able to set an SSL verify mode on faraday

taquitos:
- addresses issue #981

### DIFF
--- a/lib/octokit/default.rb
+++ b/lib/octokit/default.rb
@@ -137,6 +137,12 @@ module Octokit
         ENV['OCTOKIT_PROXY']
       end
 
+      # Default SSL verify mode from ENV
+      # @return [String]
+      def ssl_verify_mode
+        ENV['OCTOKIT_SSL_VERIFY_MODE']
+      end
+
       # Default User-Agent header string from ENV or {USER_AGENT}
       # @return [String]
       def user_agent
--- a/lib/octokit/default.rb
+++ b/lib/octokit/default.rb
@@ -138,9 +138,12 @@ module Octokit
       end
 
       # Default SSL verify mode from ENV
-      # @return [String]
+      # @return [Integer]
       def ssl_verify_mode
-        ENV['OCTOKIT_SSL_VERIFY_MODE']
+        # 0 is OpenSSL::SSL::VERIFY_NONE
+        # 1 is OpenSSL::SSL::SSL_VERIFY_PEER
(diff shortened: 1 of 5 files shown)


lib/octokit/default.rb
```diff
@@ -137,6 +137,12 @@ def proxy
         ENV['OCTOKIT_PROXY']
       end
 
+      # Default SSL verify mode from ENV
+      # @return [String]
+      def ssl_verify_mode
+        ENV['OCTOKIT_SSL_VERIFY_MODE']
```
tarebyte: So because most folks will not have this set, it will default to false which is a breaking change.

See https://github.com/lostisland/faraday/blob/923f90558c22326c9a757a7ff9dfb3a006d0abc8/lib/faraday/adapter/net_http.rb#L124-L130 for the implementation details

What do you think about doing something like this instead?

```ruby
ENV.fetch("OCTOKIT_SSL_VERIFY_MODE") { true }
```

spec/octokit/client_spec.rb
```diff
@@ -421,6 +421,13 @@
       conn = Octokit.client.send(:agent).instance_variable_get(:"@conn")
       expect(conn.proxy[:uri].to_s).to eq('http://proxy.example.com')
     end
+    it "sets an ssl verify mode" do
+      Octokit.configure do |config|
+        config.ssl_verify_mode = OpenSSL::SSL::VERIFY_NONE
+      end
+      conn = Octokit.client.send(:agent).instance_variable_get(:"@conn")
+      expect(conn.ssl[:verify_mode]).to eq(OpenSSL::SSL::VERIFY_NONE)
+    end
```
tarebyte: Can we add a test here to ensure the expected default remains unchanged?

lib/octokit/default.rb
taquitos: That makes sense, totally forgot about defaults.

spec/octokit/client_spec.rb
taquitos: Of course, good call

taquitos:
Alright, all up-to-date 🎉

KrauseFx:
So great to see this merged 🎉 Thanks!

tarebyte:

KrauseFx:
Awesome, thanks for keeping us posted 👍

PR #985, opened 2018-02-19
Add tag_name to update_release documentation

smeriwether:
The [edit_release](https://developer.github.com/v3/repos/releases/#edit-a-release) endpoint accepts a tag_name so that should be documented here.

Furthermore, when using the `update_release` method without specifying a tag_name, the tag in Github is overwritten. 

For example, my tag was "v4.0" and after calling `update_release` without a tag name it became "untagged-{commit-hash}".

### DIFF
--- a/lib/octokit/client/releases.rb
+++ b/lib/octokit/client/releases.rb
@@ -44,6 +44,7 @@ module Octokit
       # Update a release
       #
       # @param url [String] URL for the release as returned from .releases
+      # @option options [String] :tag_name Git tag from which to create release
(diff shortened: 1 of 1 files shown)


tarebyte:
Thanks for the contribution @smeriwether !!

> Furthermore, when using the update_release method without specifying a tag_name, the tag in Github is overwritten.

>For example, my tag was "v4.0" and after calling update_release without a tag name it became "untagged-{commit-hash}".

I don't _think_ that's the behavior, are you able to reproduce this using curl? It may work differently depending on whether the tag is published or unpublished.

tarebyte:

Issue #988, opened 2018-03-05
Wrong example in documentation


europ:
http://www.rubydoc.info/gems/octokit/Octokit%2FClient%2FReviews:dismiss_pull_request_review

Example with 3 arguments while 4 arguments are required.
`@client.dismiss_pull_request_review('octokit/octokit.rb', 825, 6505518)`

Missing argument:
`message (String) — The message for the pull request review dismissal`

tarebyte:
Thanks for catching that @europ! Minding opening a Pull Request?

europ:
@tarebyte Not a problem.

europ:
@tarebyte Not a problem.

PR #989, opened 2018-03-05
Command example fix in documentation.

europ:
Fix of wrong example with 3 arguments while 4 arguments are required in documentation. Example is connected to the function 'dismiss_pull_request_review'.

Closes https://github.com/octokit/octokit.rb/issues/988

\cc @tarebyte


tarebyte:

PR #996, opened 2018-03-29
Add ability to list a single deployment, by id

### LINKED ISSUE #995: Fetch a single Github Deployment by ID
From the Github API I see that it is possible to fetch a single Github Deployment (https://developer.github.com/v3/repos/deployments/#get-a-single-deployment).

That said, at frst clance it doesn't appear that octokit supports this, am I mistaken?  

It seems like this file only needs the following (https://github.com/octokit/octokit.rb/blob/master/lib/octokit/client/deployments.rb):

```ruby

      # oddly enough, this isn't defined in the upstream octokit library.
      def deployment(repo, deployment_id, options={})
        get("#{Repository.path repo}/deployments/#{deployment_id}", options)
      end
      alias :list_deployment :deployment
```

If I make a PR for this would it be accepted? Am I missing some other obvious workaround?

russellballestrini:
Fixes #995 

From the Github API I see that it is possible to fetch a single Github Deployment (https://developer.github.com/v3/repos/deployments/#get-a-single-deployment).

That said, at frst clance it doesn't appear that octokit supports this.

It seems like this file only needs the following (https://github.com/octokit/octokit.rb/blob/master/lib/octokit/client/deployments.rb):


russellballestrini:
@tarebyte could you help me figure out how to proceed? I'm hung on an error I'm not familiar with.

tarebyte:
@russellballestrini this you need to run the test locally in order to create a [VCR cassette](https://github.com/vcr/vcr). It's basically a mock that tells the test what it should expect from the API request.

tarebyte:
> I think I'm having trouble triggering VCR to create a new entry in the "cassette". Here is my PR: #996

Apologies I read this ^^ comment after I submitted my suggestion.

tarebyte:
@russellballestrini so the reason your test is having issues is because a lot of these tests are old and have been hard coded.

I took some time this morning to update the tests so that is not longer the case. Hopefully this should get this PR back on track 😄

lib/octokit/client/deployments.rb
```diff
@@ -6,6 +6,17 @@ class Client
...
+      # List a single deployment for a repository
+      #
+      # @param repo [Integer, String, Repository, Hash] A GitHub repository
+      # @param deployment_id [Integer, String, Repository, Hash] A GitHub repository
+      # @return <Sawyer::Resource> A single deployment
+      # @see https://developer.github.com/v3/repos/deployments/#get-a-single-deployment
+      def deployment(repo, deployment_id, options = {})
+        get("#{Repository.path repo}/deployments/#{deployment_id}", options)
+      end
+      alias :list_deployment :deployment
```
tarebyte: This alias doesn't really make sense, could we remove it?

russellballestrini: I could but I was following the pattern of the other methods. Does list mean "array" in this case or "look".

tarebyte: In this case list means "array"

russellballestrini: Cool! I'll fix

russellballestrini:
@tarebyte please let me know what else needs to be done for this PR.

tarebyte:
Nothing, thanks @russellballestrini !

russellballestrini:
@tarebyte would you know off chance if the team is considering a new release of `octokit.rb`? I'd like to use this new functionality. : )

tarebyte:

PR #1000, opened 2018-04-04
Call dup on arguments before mutating

greysteil:
Debugged a thorny error today: I was automatically retrying some GitHub calls in the case that I received a server error, and sometimes the retries were returning unexpected results.

With the help of @izuzak I figured out the cause was that Octokit sometimes mutates the options hashes passed to it. As a result, the retry of the request would have different params to the original request.

This PR calls `dup` before any `delete` calls are made in the client. There were already a host of methods that did this, but come had been missed.


greysteil:
@tarebyte any chance of a review on this? Imagine it's causing subtle bugs for other integrators, too!

tarebyte:
@greysteil 👋 

Thanks for the change, since this changes the behavior for Octokit (in a good way) I'm inclined to hold up until we start the process of releasing 5.0 which I plan to do next week. What do you think?

/cc @kytrinyx for thoughts

greysteil:
Works for me. I think it would be legit to ship this as a patch release, too, though - the current (mutating) behaviour isn't documented and is more likely to be interpreted as a bug than anything else I *think*.

Exciting news on 5.0! 🎉

greysteil:
(As a clarification, I mean alternatively, not in addition, to rolling into 5.0!)

kytrinyx:
We should _always_ be `dup`ing before mutating, in my opinion. I would vote to merge this into the existing master (and cut a patch release).

greysteil:
Merge in master to get the specs passing. Given the reactions on @kytrinyx's comment is this good to go?

tarebyte:

PR #1003, opened 2018-04-14
Fix pagination of GitHub App endpoints

### LINKED ISSUE #974: Pagination fails for some GitHub App APIs

rdunlop:
Fixes #974

I recently came across the same issue that @stmllr did, and it looks like he solved it in #976.

@kytrinyx asked for some small changes...so I've done those, hoping that that is sufficient.

I hope that forking/continuing the branch is an acceptable way to resolve it.


tarebyte:
This has been released in https://github.com/octokit/octokit.rb/releases/tag/v4.9.0

Issue #1004, opened 2018-04-16
Octokit can't access [bot] users by login


casperisfine:
Demo:

```ruby
>> user = Octokit.user(38352116)
=> {:login=>"shopify-shipit[bot]", :id=>38352116, :url=>"https://api.github.com/users/shopify-shipit%5Bbot%5D", ...}
>> Octokit.get(user.url)
Octokit::NotFound: GET https://api.github.com/users/shopify-shipit%255Bbot%255D: 404 - Not Found // See: https://developer.github.com/v3/users/#get-a-single-user
>> Octokit.user(user.login)
URI::InvalidURIError: bad URI(is not URI?): users/shopify-shipit[bot]
>> Octokit.user(CGI.escape(user.login))
Octokit::NotFound: GET https://api.github.com/users/shopify-shipit%255Bbot%255D: 404 - Not Found // See: https://developer.github.com/v3/users/#get-a-single-user
```

It seems Octokit doesn't know how to handle escaping properly in this case.

casperisfine:
After digging into the history, the extra escaping seem to have been introduced in https://github.com/octokit/octokit.rb/commit/f06d0763604bdf0a1f55e206a1988725c596a7b5 for the purpose of supporting the legacy search API.

Not sure if other endpoints rely on it though :/

casperisfine:
Fixed in https://github.com/octokit/octokit.rb/pull/1006

PR #1005, opened 2018-04-16
Refactor and fix path segments escaping

casperisfine:
Fix: https://github.com/octokit/octokit.rb/issues/1004

Using this branch:

```ruby
>> Octokit.user('shipit[bot]')
=> {:login=>"shipit[bot]", :id=>32029183,  ...}
>> Octokit.get("https://api.github.com/users/shipit%5Bbot%5D")
=> {:login=>"shipit[bot]", :id=>32029183, ...}
```

So there was several problems here.

First `Octokit.user('shipit[bot]')` wasn't working because `URI::Parser.new.escape` follows RFC2396 which doesn't consider `[]` as invalid characters whereas `URI.parse` follows RFC3986 which does consider `[]` as invalid characters that need to be escaped (see [MRI Issue 9990](https://bugs.ruby-lang.org/issues/9990)).

The solution here is to use `Addressable::URI.escape`, which does a proper `RFC3986` escaping. It doesn't add any dependency since `addressable` is already a faraday dependency. However it might be a problem in the future if Faraday was to drop addressable.

Then `Octokit.get("https://api.github.com/users/shipit%5Bbot%5D")` wasn't working either (even though it's an URL returned by the API) because `Connection#request` was escaping everything, which IMHO isn't a good idea.

Instead this refactor only escape specific path segments known to be able to include special characters. The downside is that I have no way to be 100% sure I haven't missed some segments which might cause regressions.

ping @tarebyte do you think this would be an acceptable patch?

### DIFF
--- a/lib/octokit/client.rb
+++ b/lib/octokit/client.rb
@@ -237,5 +237,11 @@ module Octokit
 
       conn
     end
+
+    private
+
+    def http_escape(string)
+      Addressable::URI.escape(string)
+    end
   end
 end
--- a/lib/octokit/client/labels.rb
+++ b/lib/octokit/client/labels.rb
@@ -28,7 +28,7 @@ module Octokit
       # @example Get the "V3 Addition" label from octokit/octokit.rb
       #   Octokit.labels("octokit/octokit.rb", "V3 Addition")
       def label(repo, name, options = {})
-        get "#{Repository.path repo}/labels/#{name}", options
+        get "#{Repository.path repo}/labels/#{http_escape(name)}", options
       end
 
       # Add a label to a repository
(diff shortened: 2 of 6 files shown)


tarebyte:
@casperisfine we're planning on dropping support in 5.0 for EOL rubies so if we ship these changes along with it it should be fine.

casperisfine:
Well, this issue is currently blocking me on several projects, and I can't seem to find a workaround :/

So I can try to make this work for 2.0 & 2.1 if you want.

tarebyte:
@casperisfine we're planning on doing a 5.0 ship soon, but if you can make it work for old version that would be ✨

casperisfine:
Ok, I'll see what I can do.

tarebyte:
The alternative (https://github.com/octokit/octokit.rb/pull/1006) was merged

PR #1006, opened 2018-04-18
Use Addressable::URI instead of stdlib URI for normalization

casperisfine:
Alternative to https://github.com/octokit/octokit.rb/pull/1005

Instead of trying to only escape the relevant fragments, this PR still applies escaping to all URLs, but does it with `Addressable::URI`, which is better at figuring out what needs to be escaped and what doesn't.

Example:

```ruby
>> URI::Parser.new.escape(URI::Parser.new.escape('/foo bar'))
=> "/foo%2520bar"
>> Addressable::URI.parse(Addressable::URI.parse('/foo bar').normalize.to_s).normalize.to_s
=> "/foo%20bar"
>> Addressable::URI.parse('/foo[bar]').normalize.to_s
=> "/foo%5Bbar%5D"
```

It also have the advantage of using the same parser for every MRI versions which solves the parser discrepancies in https://github.com/octokit/octokit.rb/pull/1005

@tarebyte what do you think?


casperisfine:
The `2.4.3` build ran into what look like a MRI bug. I can't trigger a retry but I think it should pass.

tarebyte:
Thanks @casperisfine!

casperisfine:
Thanks to you!

tarebyte:

PR #1009, opened 2018-04-23
Add missing remove_org_membership alias

Fryguy:
No tests added because these "org" short-form aliases are not tested in the suite.


tarebyte:
Thanks ✨

tarebyte:

PR #1010, opened 2018-05-01
[DO NOT MERGE] Spike out code generator based on octokit/routes

kytrinyx:
@tarebyte this is the spike I told you about.

/cc @gr2m @bkeepers @shiftkey @ryangribble in case you're interested :)


Rakefile
```diff
@@ -20,3 +20,12 @@ namespace :doc do
   rescue LoadError
   end
 end
+
+desc "SPIKE: Generate the API client files based on route docs."
```
kytrinyx: This is a small rake task that will read the deployments route definition from octokit/routes. It's not smart. It assumes the octokit/routes are checked out parallel to octikt.rb. It also hard-codes everything and does no error handling.

I assume that a proper rake task would read whatever route files are available and infer what client files should be generated, &c.

lib/octokit/client/deployments.rb
```diff
@@ -1,69 +1,87 @@
 module Octokit
   class Client
-
     # Methods for the Deployments API
     #
-    # @see https://developer.github.com/v3/repos/commits/deployments/
+    # @see https://developer.github.com/v3/repos/deployments/
```
kytrinyx: The old URL was just plain wrong. I checked :)

lib/octokit/client/deployments.rb
```diff
@@ -1,69 +1,87 @@
...
     module Deployments
-
-      # Fetch a single deployment for a repository
+      # Get a single deployment
```
kytrinyx: The new documentation comes directly from the route file, which is generated from the docs site. I'd prefer to be consistent with that documentation, even if it's slightly less descriptive than the old stuff.

lib/octokit/client/deployments.rb
```diff
@@ -1,69 +1,87 @@
...
       #
       # @param repo [Integer, String, Repository, Hash] A GitHub repository
-      # @param deployment_id [Integer, String, Repository, Hash] A GitHub repository
+      # @param deployment_id [Integer] The ID of the deployment
```
kytrinyx: This line was just plain wrong, as well.

The routes file's description for this field is an empty string, since it's not described in the docs site (this is a field that is part of the URL). I made a fallback description for `*_id` fields that are empty "the ID of the _$whatever_".

lib/octokit/client/deployments.rb
```diff
@@ -1,69 +1,87 @@
...
       end
 
-      # List all deployments for a repository
+      # List deployments
       #
       # @param repo [Integer, String, Repository, Hash] A GitHub repository
+      # @param options [String] :sha The SHA that was recorded at creation time.
+      # @param options [String] :ref The name of the ref. This can be a branch, tag, or SHA.
+      # @param options [String] :task The name of the task for the deployment (e.g., `deploy` or `deploy:migrations`).
+      # @param options [String] :environment The name of the environment that was deployed to (e.g., `staging` or `production`).
```
kytrinyx: One of the things that we sometimes do is list each of the optional params

```
@param options [type] :name_of_param $DESCRIPTION`
```

I really like that. A corollary is that if we have no optional params, we can leave off the `@param options` altogether.

lib/octokit/client/deployments.rb
```diff
@@ -1,69 +1,87 @@
...
+      # @param options [Array<string>] :required_contexts The status contexts to verify against commit status checks. If this parameter is omitted, then all unique contexts will be verified before a deployment is created. To bypass checking entirely pass an empty array. Defaults to all unique contexts.
+      # @param options [String] :payload JSON payload with extra information about the deployment.
+      # @param options [String] :environment Name for the target deployment environment (e.g., `production`, `staging`, `qa`).
+      # @param options [String] :description Short description of the deployment.
+      # @param options [Boolean] :transient_environment Specifies if the given environment is specific to the deployment and will no longer exist at some point in the future. **This parameter requires a custom media type to be specified.**
+      # @param options [Boolean] :production_environment Specifies if the given environment is one that end-users directly interact with. **This parameter requires a custom media type to be specified.**
```
kytrinyx: This looks like a big block of changes but mostly it's two things:

- the descriptions changed
- there are some additional optional arguments that weren't listed before.

The arguments that are under preview had some kind of weird (in this context) descriptions that said "see note below". I hard-coded a fix to remove that.

lib/octokit/client/deployments.rb
```diff
@@ -1,69 +1,87 @@
...
+      # @return <Sawyer::Resource> The new deployment
       # @see https://developer.github.com/v3/repos/deployments/#create-a-deployment
       def create_deployment(repo, ref, options = {})
         options[:ref] = ref
         post("#{Repository.path repo}/deployments", options)
       end
 
-      # List all statuses for a Deployment
+      # Get a single deployment status
```
kytrinyx: This endpoint wasn't supported in the octokit at all. Bonus!

That said... we should probably also generate tests, which I've not done here.

lib/octokit/client/deployments.rb
```diff
@@ -1,69 +1,87 @@
...
+      # List deployment statuses
       #
-      # @param deployment_url [String] A URL for a deployment resource
+      # @param deployment_url [String] A URL for a deployment resource.
       # @return [Array<Sawyer::Resource>] A list of deployment statuses
       # @see https://developer.github.com/v3/repos/deployments/#list-deployment-statuses
       def deployment_statuses(deployment_url, options = {})
-        deployment = get(deployment_url, :accept => options[:accept])
+        deployment = get(deployment_url, accept: options[:accept])
         get(deployment.rels[:statuses].href, options)
```
kytrinyx: This one was a bit tricky since it uses the rel rather than putting together the URL from the parameters defined in the routes file. In other words, I have done gross, gross things in order to support this. The TL;DR of the grossness is: (a) I override the params to add the `deployment_url` and delete the ID param and the repo param, and (b) I have a conditional that switches on whether it's a subresource or not, to determine what sort of method implementation to generate. I am almost 100% certain that this is going to turn out to not be good enough, because I would bet that we are not consistent in our choice of implementation. But for now this works.

lib/octokit/client/deployments.rb
```diff
@@ -1,69 +1,87 @@
...
       end
       alias :list_deployment_statuses :deployment_statuses
```
kytrinyx: We have aliases for all the list endpoints.

lib/octokit/client/deployments.rb
```diff
@@ -1,69 +1,87 @@
...
+      # @param options [String] :description A short description of the status. Maximum length of 140 characters.
+      # @param options [String] :environment_url Sets the URL for accessing your environment. **This parameter requires a custom media type to be specified.**
+      # @param options [Boolean] :auto_inactive Adds a new `inactive` status to all non-transient, non-production environment deployments with the same repository and environment name as the created status's deployment. **This parameter requires a custom media type to be specified.**
+      # @return <Sawyer::Resource> The new deployment status
       # @see https://developer.github.com/v3/repos/deployments/#create-a-deployment-status
       def create_deployment_status(deployment_url, state, options = {})
-        deployment = get(deployment_url, :accept => options[:accept])
+        deployment = get(deployment_url, accept: options[:accept])
         options[:state] = state.to_s.downcase
```
kytrinyx: Enums are going to be interesting. For now I assume that they should be normalized by downcasing them. This might not actually be true.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
+require 'forwardable'
+require 'json'
+require 'pathname'
+require 'active_support/inflector'
+
+module Spike
+  class Resource
```
kytrinyx: This is a small class that tries to figure out what resource we're dealing with based on the path.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+
+      @objects = segments.reject {|segment| segment.start_with?(":")}
+      if @objects.count > 1
+        @objects = @objects.reject {|resource| resource == "repos"}
+      end
+      @objects
+    end
+  end
+
+  class Routes
```
kytrinyx: This is a small class that, given the route definition from octokit routes, will override and normalize to make it easier to generate what we need later.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+    def initialize(json)
+      @definition = JSON.parse(json, object_class: OpenStruct)
+      normalize!
+    end
+
+    private
+
+    def normalize!
+      @definition["original_path"] = definition.path
+      @definition["path"]          = normalized_path
```
kytrinyx: original vs actual path -- for now this is whenever we have something in the path that says `:id` we need it to be prefixed with the actual resource name (singular).

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+      @definition["params"]        = normalized_params
+
+      @definition["documentation_url"] = @definition["documentationUrl"]
```
kytrinyx: The routes file uses `camelCase`, I don't want to have to deal with that in the rest of the code.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+      @definition["verb"]              = @definition["method"]
```
kytrinyx: This works around the problem with the `Object#method` method by calling it `verb` instead  so we get `definition.verb`.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+        resource = original_path_segments[idx-1].singularize
+        path = path.gsub(":id", ":#{resource}_id")
+      end
+      path
+    end
+
+    def normalized_params
+      params = (@definition.params || []).reject do |param|
+        param.name == "owner"
+      end
```
kytrinyx: Get rid of `owner` since we have a really fancy way of doing the `repo`.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+
+      params.each do |param|
+        if param.name == "id"
+          idx = original_path_segments.index(":id")
+          resource = original_path_segments[idx-1].singularize
+          param["name"] = "#{resource}_id"
```
kytrinyx: Same as normalizing the path, we need to normalize the param that will fill that slot in the URL.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+        end
+        if param.name.end_with?("_id") && param.type == "string"
+          param["type"] = "integer"
+        end
```
kytrinyx: The routes file says that these are strings, but they're not.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+        if param.name == "repo"
+          param["description"] = "A GitHub repository"
+        end
```
kytrinyx: The routes file has an empty string as the description for this, since it doesn't come from the table of inputs on the developer site, it's inferred from the API endpoint path.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+        if param.description == "" && param.name.end_with?("_id")
+          param["description"] = "The ID of the #{param.name.gsub("_id", "").gsub("_", " ")}"
```
kytrinyx: This is where I add a default description for any ID params that don't already have one.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+        end
+        param.description = param.description.to_s.gsub(" Please see more in the alert below.", "")
```
kytrinyx: Here's where we ditch the bit about an alert below, which we don't have here for obvious reasons.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+        if param.type.to_s.end_with?("[]")
+          param["type"] = "Array<#{param.type.gsub("[]", "")}>"
```
kytrinyx: Convert array types since the routes file doesn't match the syntax we use in the tomdoc.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+        end
+      end
+
+      resource = Spike::Resource.new(definition.path)
+      if resource.subresource?
```
kytrinyx: If it's a subresource we want a `thingy_url` rather than the repo and id param of the parent resource.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+      def parameterize(args)
+        "#{args.join(", ")}, options = {}"
+      end
+    end
+
+    class KwargsParameterizer
+      def parameterize(args)
+        "#{args.map {|arg| arg + ":"}.join(", ")}, **options"
+      end
+    end
```
kytrinyx: This generator supports two versions of the method parameters. I want us to be able to switch easily between them.

I used `options = {}` in the positional one only to minimize the diff on the actual code implementation. We can change that later, of course, if we decide we'd rather use the hash splat.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+
+    extend Forwardable
```
kytrinyx: This along with the `delegate` line on 144 is how to delegate in regular ruby without rails.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+
+    VERB_PRIORITY = %w(GET POST)
```
kytrinyx: I want to sort the client methods consistently throughout all the files:

* get a single thing
* list the things
* create a thing

and eventually we'll get the "update and delete" when we spike out another API that has them.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+        end
+      when "POST"
+        "The new #{resource.name.singularize.gsub("_", " ")}"
+      else
+      end
+    end
+
+    def verb
+      definition["method"]
+    end
```
kytrinyx: Oh. I seem to have accidentally left this in. Can be ✂️

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+
+    def parameters
+      parameterizer.parameterize(arguments)
+    end
```
kytrinyx: This is where we use the positional or kwarg version of the parameters.

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+
+    def arguments
+      definition.params.select(&:required).map(&:name)
```
kytrinyx: I had been programming for several years before I realized the difference between parameters (part of the method signature) and arguments (the inputs when you call that method).

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+      case verb
+      when "GET"
+        resource.name
+      when "POST"
+        "create_#{resource.name.singularize}"
+      else
+      end
+    end
+
+    def alternate_name
```
kytrinyx: This is for the alias. But good luck naming a method `alias` :)

lib/spike.rb
```diff
@@ -0,0 +1,361 @@
...
+      "list_#{resource.name}"
+    end
+
+    def parts
+      @parts ||= path_segments[path_segments.index(directory)..-1].reject {|segment| segment.start_with?(":")}
+    end
+
+    def priority
+      [parts.count, VERB_PRIORITY.index(verb), resource.singular?? 0 : 1]
+    end
```
kytrinyx: This is so we can say `sort_by(&:priority)` on the endpoints to get the method definitions in the right order.

spec/spike_spec.rb
```diff
@@ -0,0 +1,147 @@
+require_relative '../lib/spike'
+
+# TODO:
+# - for a subresource normalize route params to use url instead of repo + base resource
+# - fix @return (right now it's inverted to check for singular, not collection)
```
kytrinyx: Oh. I did both of these. ✂️

spec/spike_spec.rb
```diff
@@ -0,0 +1,147 @@
...
+
+describe Spike do
```
kytrinyx: The tests are not extensive, just enough to help implement the spike.

kytrinyx:
Gah. Failing for `endpoints.first&.documentation_url` of course.

lib/octokit/client/deployments.rb
gr2m: I think the docs team is also happy to accept small pull requests that improve the descriptions, this way all the Octokit REST clients will benefit, too :)

lib/spike.rb
ryangribble: Is the routes file incorrect and should be fixed?

kytrinyx: Yeah, maybe. I mean—they occur within the URL path so I can also see how they could be seen as strings, but in payloads when they are delivered they are ints. @gr2m what do you think?

PR #1012, opened 2018-05-09
Document that PR review endpoints use PR number, not PR ID

greysteil:
The use of `id` as a variable name here is a bit confusing - using `number` makes it more obvious what's going on.

Docs [here](https://developer.github.com/v3/pulls/review_requests/).


tarebyte:
🍰

Issue #888, opened 2017-05-09
Retries are not consistently applied on API errors


ohookins:
It seems that the repo statistics methods of the client support retrying when there is an error or timeout from the API:
https://github.com/octokit/octokit.rb/blob/836948b6c176d7829e250e65caea790721c96523/lib/octokit/client/stats.rb#L91-L101

It would be very useful to have this universally applied throughout the SDK on the client itself (rather than per endpoint) so that the application doesn't have to implement its own retry logic.

kytrinyx:
@ohookins that's a good point, this is definitely inconsistent.

I think that having retry logic within the calls to the individual endpoints is probably not the right approach for the client, but it would be very useful to have helpers to make it easy for people to add retry logic themselves if they need/want it.

ohookins:
@kytrinyx I can appreciate that approach. In practice, with something like the Retryable gem, I would say that the experience is less than optimal. I feel that the AWS Ruby SDK does a very good job of encapsulating this logic though, if that's something you could look into.

kytrinyx:
Thank you, I will definitely look into that one.

kytrinyx:
We've discussed this, and we've decided that we want to add functionality that makes it easy to do retries, however we won't be adding it automatically to all the endpoints.

AbNoir:
rery_timeout = options.delete ( : retry_timeout )
   retry_wait = options.delete ( : retry_wait ) ||  0.5 
   مهلة =  الوقت . الآن + retry_timeout
 النهاية 
 حلقة  القيام به 
   البيانات = الحصول على ( " # { مستودع .path الريبو } / احصائيات / # { قياس } " ، والخيارات)
   العودة البيانات إذا last_response.status ==  200 
   العودة  شيء  ما لم retry_timeout
   العودة  لا شيء  إذا  الوقت . الآن > = المهلة
   سليب retry_wait إذا retry_wait
 النهاية

ohookins:
I haven't done anything about this on my end, although I was looking into making a Faraday middleware to implement the retries. Perhaps that's an acceptable alternative without introducing a new concern into the SDK? But it might help to have an example implementation somewhere.

kytrinyx:
> it might help to have an example implementation somewhere.

Yeah, definitely. If you do get something working at the middleware level, would you post something here?

kytrinyx:
> Seems the "fix" to this issue is simply to add it to the README default config?

That is my favorite kind of fix :) Looking at your PR now.

[...]

PR #1021, opened 2018-05-29
Faraday retry middleware

tjoyal:
Ref https://github.com/octokit/octokit.rb/issues/888

I think that is all the required changes to make everyone happy regarding intermittent Github 5xx error class errors that could be easily replayed.

🎩manually with the `Octokit::NotFound` error. 

Thoughts? I'm not super familiar with Octokit (it just work so I never had to look inside)

### DIFF
--- a/README.md
+++ b/README.md
@@ -549,6 +549,7 @@ traffic:
 
 ```ruby
 stack = Faraday::RackBuilder.new do |builder|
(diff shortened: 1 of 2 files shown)


kytrinyx:
> it just work so I never had to look inside

🌷 🌷 🌷 So happy it mostly just works!

This looks great, thank you!

abinoda:
@tjoyal Thanks for this update. Should the docs be updated? How does one enable this?

PR #1025, opened 2018-06-12
Fix contributors endpoint for empty repos

d12:
Fixes https://github.com/octokit/octokit.rb/issues/912
Closes https://github.com/octokit/octokit.rb/pull/994

Me, @srinjoym and @benemdon finished up @Kadaaran 's work from #994, fixing the timeout bug in the contributors endpoint.

GitHub returns a 204 when asking for statistics on empty repos. Previously, Octokit would continue to retry until the retry timeout was reached.

This PR just returns `[]` when a 204 is returned.

Also wrote a few tests showing we don't retry when a 204 is returned, and that octokit returns `[]` when GitHub returns a 204.

cc @tarebyte @kytrinyx for review

### DIFF
--- a/lib/octokit/client/stats.rb
+++ b/lib/octokit/client/stats.rb
@@ -96,7 +96,7 @@ module Octokit
         loop do
           data = get("#{Repository.path repo}/stats/#{metric}", options)
           return data if last_response.status == 200
-          return 0 if last_response.status == 204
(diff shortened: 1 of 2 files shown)


d12:
Friendly bump @kytrinyx :)

d12:
bump @tarebyte @kytrinyx

Issue #893, opened 2017-05-25
Using the methods directly on `Octokit` are not thread safe


wagenet:
When using methods directly on the `Octokit` module, the client is shared across all threads. This would be ok, except for the fact that the client maintains state. One place this happens in `last_response` which, if two threads are making requests at once, could easily end up out of order. If an explicit instance of `Octokit::Client` is created, this problem can be avoided.

See https://gist.github.com/chancancode/0061e8972b8e738b3734ceef6fe5ab23#file-octokit_threads-rb for a reproduction. Run with `GLOBAL=1` to see it fail. (Note that it can take a minute or so to see output from the script.)

My recommendation would be to deprecate use of methods directly on `Octokit`. Alternatively, there probably are some other routes you could take to ensure that the client is per-thread even when using the `Octokit` module directly.

/cc @chancancode

chancancode:
To be more precise for the benefits of future readers – the client itself is still not threadsafe, as in if you make a client and hand that to multiple threads you will still run into the same problem. However, if you can make each thread *own* one client then as far as I can tell there aren't other obvious global shared sates floating around, which would match most people's use cases and expectations IMO

kytrinyx:
> My recommendation would be to deprecate use of methods directly on Octokit

Yes, we'll go ahead and deprecate the use of those methods directly, and add a warning that the client is not threadsafe and that if people are using threads, they should be sure to make a new client for each thread.

chancancode:
@kytrinyx thank you!

abinoda:
@kytrinyx As part of this, do you think the README should be updated to not use examples where methods are being called on `Octokit` directly? Want me to open a PR for this?

PR #1026, opened 2018-06-12
Updates README to use instance methods

abinoda:
This is related to https://github.com/octokit/octokit.rb/issues/893.

The purpose of this PR is to update the README so it does not encourage developers to use methods directly on the Octokit module by updating example code to use Octokit client instances instead. Using methods directly on the Octokit module is not thread-safe (I've run into this firsthand), and @kytrinyx suggested that this ability would be deprecated. 

I also fixed a coding style inconsistency was bugging me where sometimes we had:
```ruby
client = Octokit::Client.new(:login => 'defunkt', :password => 'c0d3b4ssssss!')
```

and in other places:
```ruby
client = Octokit::Client.new \
  :login    => 'defunkt',
  :password => 'c0d3b4ssssss!'
```

I feel like the former is what we should be encouraging to new developers so I converted everything over to that format.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,10 +54,11 @@ Most methods have positional arguments for required input and an options hash
 for optional parameters, headers, or other options:
 
 ```ruby
-# Fetch a README with Accept header for HTML format
-Octokit.readme 'al3x/sovereign', :accept => 'application/vnd.github.html'
-```
+client = Octokit::Client.new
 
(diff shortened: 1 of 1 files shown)


abinoda:
@tarebyte Was there any follow-up needed from me on this?

tarebyte:
Nope! For some reason I approved it and forgot to merge it 😅

PR #978, opened 2018-01-17
Add "List all topics for a repository" and "Replace all topics for a repository"

jpmckinney:
replaces #977, closes #933

PR #1044, opened 2018-08-06
Implements Topics API

### LINKED ISSUE #933: Add support for topics endpoints
These are now available via the REST API.

- https://developer.github.com/v3/repos/#list-all-topics-for-a-repository
- https://developer.github.com/v3/repos/#replace-all-topics-for-a-repository

JuanitoFatas:
This Pull Request implements [Repositories REST API V3](https://developer.github.com/v3/repos/) [list all topics][list-all] and [replace all topics][replace-all] endpoints.

Response for `topics`:

```ruby
> client.topics "octokit/octokit.rb"
{:names=>["octokit", "github", "github-api"]} # A Sawyer::Resource instance
``` 

Response for `replace_all_topics`:

```ruby
> client.replace_all_topics "juanitofatas/octokit.rb", ["octokit", "github", "github-api"]
{:names=>["github", "octokit", "github-api"]} # A Sawyer::Resource instance
``` 

Closes #978 Closes #933

[list-all]: https://developer.github.com/v3/repos/#list-all-topics-for-a-repository
[replace-all]: https://developer.github.com/v3/repos/#replace-all-topics-for-a-repository


JuanitoFatas:
@kytrinyx @tarebyte Please review 👀:pray:

spec/octokit/client/repositories_spec.rb
```diff
@@ -273,6 +273,26 @@
     end
   end # .collaborators
 
+  describe ".topics", :vcr do
+    it "returns repository topics" do
+      topics = Octokit.topics("github/linguist", :accept => 'application/vnd.github.mercy-preview+json')
```
tarebyte: Instead of using the string, could we use the value from the `PREVIEW_TYPES` hash?
https://github.com/octokit/octokit.rb/blob/master/lib/octokit/preview.rb#L19

spec/octokit/client/repositories_spec.rb
```diff
@@ -273,6 +273,26 @@
     end
   end # .collaborators
 
+  describe ".topics", :vcr do
```
tarebyte: Could we put these under the `with repository` context?

That way if we ever have to re record these cassettes another user can run them with their own data?

lib/octokit/client/repositories.rb
```diff
@@ -380,6 +380,39 @@ def repository_teams(repo, options = {})
       alias :repo_teams :repository_teams
       alias :teams :repository_teams
 
+      # List all topics for a repository
+      #
+      # Requires authenticated client for private repos.
+      #
+      # @param repo [Integer, String, Hash, Repository] A GitHub repository.
+      # @return [Array<Sawyer::Resource>] Hash of array representing topics.
```
jpmckinney: Do you mean "Array of hashes"?

lib/octokit/client/repositories.rb
```diff
@@ -380,6 +380,39 @@ def repository_teams(repo, options = {})
...
+        opts = ensure_api_media_type(:topics, options)
+        paginate "#{Repository.path repo}/topics", opts
+      end
+
+      # Replace all topics for a repository
+      #
+      # Requires authenticated client.
+      #
+      # @param repo [Integer, String, Repository, Hash] A Github repository
+      # @param names [Integer] Number ID of the issue
```
jpmckinney: `names` is not "Number ID of the issue"

lib/octokit/client/repositories.rb
```diff
@@ -380,6 +380,39 @@ def repository_teams(repo, options = {})
...
+      # @return [Array<Sawyer::Resource>] A list of the topics currently on the issue
```
jpmckinney: Wrong docstring for `@return`.

spec/octokit/client/repositories_spec.rb
JuanitoFatas: Good idea 👍 

I am trying (by the way, there are two `with repository` contexts) and what [test ENVs](https://github.com/octokit/octokit.rb#running-and-writing-new-tests) do I need in order to record cassettes correctly.

It's getting closer to my sleep time. If you could show me how to move `topics` spec inside one of the context will be very helpful. I can update the other one tomorrow.

lib/octokit/client/repositories.rb
jpmckinney: Aha, then `[Array<Sawyer::Resource>]` is what needs to change.

JuanitoFatas: I already changed in https://github.com/octokit/octokit.rb/pull/1044/commits/f927472fb879912d4dde80ece2ce194f762c2ce9, thank you.

JuanitoFatas:
@tarebyte I've moved tests into with a repository context ✨

spec/octokit/client/repositories_spec.rb
tarebyte: Apologies for not getting back to this, glad you figured it out!

spec/octokit/client/repositories_spec.rb
```diff
@@ -200,6 +200,26 @@
...
+        expect(topics.names).to include("octokit")
+        assert_requested :get, github_url("/repos/#{@repo.full_name}/topics")
+      end
+    end # .topics    
+
+    describe ".replace_all_topics", :vcr do
+      it "replaces all topics for a repository" do
+        new_topics = ["octocat", "github", "github-api"]
+        options = {
+          :accept => "application/vnd.github.mercy-preview+json"
```
kytrinyx: I missed this the first time around. Would you please access this through the `PREVIEW_TYPES` constant?

JuanitoFatas: Good catch!

tarebyte:
Thanks @JuanitoFatas!!

PR #1050, opened 2018-08-15
Add in note about require

andymckay:
Just in case, like me you forgot about it in a moment of stupidity.


greysteil:
My change in #1055 will fix the specs here.

PR #1054, opened 2018-08-16
Remove the :name key from #add_team_repository request

### LINKED ISSUE #1051: add_team_repo call now gets 422 from GitHub for "name" is not a permitted key.
Line of code that I think is driving this https://github.com/octokit/octokit.rb/blame/master/lib/octokit/client/organizations.rb#L527

Reference to the api itself https://developer.github.com/v3/teams/#add-or-update-team-repository - permission is only allowed field to be passed. 

My assumption is that prior to today; GitHub did not fail when it received additional parameters; and something changed on their end. I now get this error when calling the add_team_repo function: 

```
"name" is not a permitted key. // See: https://developer.github.com/v3/teams/#add-or-update-team-repository
```

I'm uncertain how the options.merge function works; however I assume you could run that, and then delete the name param before passing it back to GitHub. 

I know code contributions are much more appreciated than issue reporting :P - but I have not yet figured out my company's policy on open source contributions.

BenEmdon:
## What
It appears the API began validating the presence of parameters on the request `teams/:team_id/repos/:org/:repo_name`. When a request contained the parameter key `name` it would be rejected by the API with a status of `422`. 

Using the method `#add_team_repository` would yield a `422` status with the message:
```js
"name" is not a permitted key. // See: https://developer.github.com/v3/teams/#add-or-update-team-repository
```

This PR proposes a fix to drop the extra param `:name` which no longer serves a purpose.

API reference [`https://developer.github.com/v3/teams/#add-or-update-team-repository`](https://developer.github.com/v3/teams/#add-or-update-team-repository)

---
Fixes #1051 closes #1053

### DIFF
--- a/lib/octokit/client/organizations.rb
+++ b/lib/octokit/client/organizations.rb
@@ -524,7 +524,7 @@ module Octokit
       # @example Add a team with admin permissions
       #   @client.add_team_repository(100000, 'github/developer.github.com', permission: 'admin')
       def add_team_repository(team_id, repo, options = {})
-        boolean_from_response :put, "teams/#{team_id}/repos/#{Repository.new(repo)}", options.merge(:name => Repository.new(repo))
+        boolean_from_response :put, "teams/#{team_id}/repos/#{Repository.new(repo)}", options
       end
       alias :add_team_repo :add_team_repository
 


d12:
@kytrinyx Do you think you could take a peek at this when you get a moment?

BenEmdon:
To prove that this method is broken I made a PR with a failing test, then patched classroom with my octokit.rb fork.

See https://github.com/education/classroom/pull/1483

kytrinyx:
This is my bad. I meant to deploy this validation as a scientist experiment to see what would fail without actually making things fail. I will deploy a fix momentarily.

BenEmdon:
@kytrinyx no problem, thanks for the fix upstream!
Should this patch still merge if it is equivalent to the old way (but fitting within the ideals of upstream)?

kytrinyx:
I think we should merge the patch. The API doesn't accept a `:name` parameter according to the docs:
https://developer.github.com/v3/teams/#add-or-update-team-repository

The codebase just ignores it currently, (well, it did until two days ago /facepalm). So while we should make the API backwards compatible, it would make sense to update the client to actually match the documented behavior.

PR #1055, opened 2018-08-17
Use new app installation token endpoint

greysteil:
The app installation token routes is being moved, as publicised in [this changelog entry](https://developer.github.com/changes/2018-08-16-renaming-and-deprecation-of-github-app-installation-access-token-route/).

This PR updates the endpoint Octokit uses to the new one.


greysteil:
@tarebyte - looks like you're the right person to review this. Would it be possible to get a speedy release if you're happy with it? That September 4th deadline for removing the route is soon!

alienxp03:
Thanks for the PR. Just tried the gem and it failed to install the app. Didn't realized there's a new endpoint for the app installation.

abinoda:
@tarebyte @kytrinyx I wanted to echo @greysteil's request for this to be released as soon as possible so there's time to integrate it before the breaking change in the API on September 4th – that date is coming up 😨

PR #1060, opened 2018-09-05
Add support for Transfer repository API

Yuji-Kuroko:
This PR add support for Transfer a repository API.
https://developer.github.com/v3/repos/#transfer-a-repository


lib/octokit/client/repositories.rb
```diff
@@ -177,6 +177,19 @@ def delete_repository(repo, options = {})
...
       alias :delete_repo :delete_repository
 
+      # Transfer repository
+      #
+      # Requires authenticated repository or organization owner of source and destination.
+      #
+      # @see https://developer.github.com/v3/repos/#transfer-a-repository
+      # @param repo [Integer, String, Hash, Repository] A GitHub repository
+      # @return [Sawyer::Resource] Repository info for the transferred repository
+      def transfer_repository(repo, options = {})
```
kytrinyx: @tarebyte do you think we should make `new_owner` a positional argument here, since it is required?

lib/octokit/client/repositories.rb
```diff
@@ -177,6 +177,19 @@ def delete_repository(repo, options = {})
       end
       alias :delete_repo :delete_repository
 
+      # Transfer repository
+      #
+      # Requires authenticated repository or organization owner of source and destination.
+      #
+      # @see https://developer.github.com/v3/repos/#transfer-a-repository
+      # @param repo [Integer, String, Hash, Repository] A GitHub repository
+      # @return [Sawyer::Resource] Repository info for the transferred repository
```
kytrinyx: We should document the optional parameters as well.

Take a look at https://github.com/octokit/octokit.rb/blob/f72f466d6b183d24b461a7c2cbf1c5cbec22ac3d/lib/octokit/client/authorizations.rb#L44-L49 to see how that's done elsewhere.

lib/octokit/client/repositories.rb
```diff
@@ -177,6 +177,19 @@ def delete_repository(repo, options = {})
       end
       alias :delete_repo :delete_repository
 
+      # Transfer repository
+      #
+      # Requires authenticated repository or organization owner of source and destination.
```
kytrinyx: I'm not sure I understand what this comment is saying. Would you clarify, please?

spec/octokit/client/repositories_spec.rb
```diff
@@ -484,6 +484,24 @@
     end
   end # .repository?
 
+  describe ".transfer_repository", :vcr do
+    it "repository transfer from myself to my organization" do
+      repository = @client.create_repository("an-repo", auto_init: true)
```
kytrinyx: Nit: `s/an-repo/a-repo`

spec/octokit/client/repositories_spec.rb
```diff
@@ -484,6 +484,24 @@
...
+      accept_header = Octokit::Preview::PREVIEW_TYPES[:transfer_repository]
+      @client.transfer_repository(repository.full_name, { new_owner: test_github_org, accept: accept_header })
+      assert_requested :post, github_url("/repos/#{repository.full_name}/transfer")
+
+      result = @client.repository?("#{test_github_org}/#{repository.name}")
+      expect(result).to be true
+
+      # cleanup
```
kytrinyx: Since this is in a separate `describe` we can use a separate `before` block for the setup and an `after` block for the cleanup.

lib/octokit/client/repositories.rb
tarebyte: @kytrinyx yup great idea 👍

lib/octokit/client/repositories.rb
Yuji-Kuroko: Thank you. I will add the document.

lib/octokit/client/repositories.rb
Yuji-Kuroko: Oh, I misunderstand that.
I tried transfer from user to organization.
And I saw this [docs](https://help.github.com/articles/transferring-a-repository-owned-by-your-organization/).
What kind of document should I write?

Yuji-Kuroko:
@kytrinyx Thank you for your polite review!
I fixed some reviewed points.
Please check it.

Yuji-Kuroko:
Travis CI failed only ruby 2.2.10.
I will rebuild Travis CI.
(I will close and reopen this PR for Travis CI rebuild trigger.)

lib/octokit/client/repositories.rb
kytrinyx: How about just replacing the first line of the comment with:

> Transfer a repository owned by your organization

And then getting rid of the "requires authenticated...". If we link to the docs, then they can see the documentation in detail by following the link.

kytrinyx:
This is looking great. I have one more suggestion for the documentation tweak, and then I think we can pull this in.

PR #1083, opened 2018-10-12
[WIP] Checks api

jfragoulis:
I have began implementing the Checks API. Could you take a look and let me know if I am on the correct path? It is work in progress and I do **not** expect a full review, only guidelines for this I might have missed.


tarebyte:
@jfragoulis this is the right direction, please continue!

mqchau:
Thank you for doing this. Looking forwards to try it out

jfragoulis:
There are differences between the [.com](https://developer.github.com/v3/checks/runs/#create-a-check-run) and [enterprise](https://developer.github.com/enterprise/2.14/v3/checks/runs/#create-a-check-run) docs (For example under check run annotations).

How should I go about this?

tarebyte:
@jfragoulis would you mind pointing out the differences? I'm skimmed through and I'm not quite sure I see what you're talking about.

jfragoulis:
In check run annotaions' documentation, there are a couple of differences between the enterprise and the .com. I am working with the enterprise, so maybe the .com documentation is out of date, maybe the implementation is different, but for example, some differences I found are:

| Enterpise | .com |
| --- | --- | 
| `filename` | `path` |
| `blob_href` | none |
| `warning_level` | `annotation_level` |

Maybe it does not matter, maybe it does. If someone passes an unsupported key, the api will return the error anyway. 
But which version should I choose to document the octokit.rb method?

jfragoulis:
@tarebyte I need some guidance for the specs. I have been struggling for hours but apart from the fact that I was not familiar with the setup (I am now a bit), it is also all new ground (installation setup and authentication) and VCR is also giving a headache.

What I am trying to do is have a simple test pass to begin with, for example: `returns a single check suite`.

I have managed to setup an authenticated app, as seen in the `apps_spec.rb`:

```ruby
before do 
  Octokit.reset!
  @jwt_client = Octokit::Client.new(bearer_token: new_jwt_token)
  use_vcr_placeholder_for(@jwt_client.bearer_token, '<JWT_BEARER_TOKEN>')
end
```

Now, to test the checks api, I need to setup a client for an authorized installation. I have the app and I need to install it to a repo (the `@test_repo` I guess), then create an installation access token and create the client.

Now, I have see the `test_github_integration_installation` in the helpers, so I know there is an installation already setup. Is it installed on the `@test_repo` tho?

Also, my biggest problem is VCR. I am always getting the `An HTTP request has been made that VCR does not know how to handle` error. 

I have tried doing:

```ruby
VCR.use_cassette("Octokit_Client_Apps/with_app_installation/_create_app_installation_access_token/creates_an_access_token_for_the_installation") do
 ...
end
```

in order to create the installation token. I _think_ it worked.

I have pushed a [wip](https://github.com/octokit/octokit.rb/pull/1083/commits/14225ce62d948c9d432b3be4bf5d74024c048172) commit with what I am trying so far.

jfragoulis:
Is anyone there? I could use some help.

lib/octokit/client/check_runs.rb
```diff
@@ -0,0 +1,160 @@
...
+      # @param id [Integer] The check suite id
+      # @param options [Hash] Method options
+      # @option options [String] :check_name Returns check runs with the specified name.
+      # @option options [String] :status Returns check runs with the specified status. Can be one of queued, in_progress, or completed.
+      # @option options [String] :filter Filters check runs by their completed_at timestamp. Can be one of latest (returning the most recent check runs) or all. Default: latest
+      # @return [Array<Sawyer::Resource>] Array of check runs
+      # @see https://developer.github.com/v3/checks/runs/#list-check-runs-in-a-check-suite
+      # @example
+      #   Octokit.check_runs('rails/rails', 1, state: 'closed')
+      def check_runs(repo, id, options = {})
```
timoschilling: this method overwrites the method from line 111!

laserlemon:
@tarebyte, implemented by #1101.

tarebyte:
@jfragoulis I'm sorry I let this drop, I'm closing this as another pull request came in that got merged https://github.com/octokit/octokit.rb/pull/1101

PR #1090, opened 2018-11-20
Support auto pagination in org_projects

mattr-:
This allows the `auto_paginate` setting on the Client object to work with `org_projects` and prevents people from having to paginate manually even when using auto pagination.


adamdiel:
*bump* This would be great to have !!

adamdiel:
@tarebyte - is something wrong with the build node ... a quick glance was showing that bundle wasn't installed.

tarebyte:
Yeah, it's because we have a wide variety of ruby versions.

PR #1092, opened 2018-11-29
Fix docs of #repository? return value type

vecerek:
#434 ensured the predicate method to return a boolean but forgot to update the return tag.


tarebyte:
🍰

PR #1094, opened 2018-12-22
Change the `raise_invalid_repository!` msg from `repo/user` to `user/repo`

DaiAoki:
I found an mistake of error message in `raise_invalid_repository!` method.
`repo/user` -> `user/repo`

I'm being helped by you, octokit.rb!


markaw67:
What exactly is this change going to do.

PR #1099, opened 2019-01-10
Update ruby support

tnaoto:
Hi

With the release of bundler v2, the CI of ruby ​​2.3 was broken.

https://bundler.io/blog/2019/01/03/announcing-bundler-2.html

In order to fix this we need to use bundler v2, and as a result we need to finish support before ruby ​​2.2.
Since Ruby 2.2 is EOL, I think that there is no problem.

Thanx.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,6 @@ language: ruby
 
 rvm:
   - jruby-19mode
-  - 2.0
-  - 2.1
-  - 2.2.10
   - 2.3.7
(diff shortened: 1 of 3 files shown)


tarebyte:
@tnaoto it looks like coveralls is having issues reporting, mind checking it out?

Thanks!

tarebyte:
Closing out as this was fixed in https://github.com/octokit/octokit.rb/pull/1101

PR #1106, opened 2019-02-14
Add `Octokit::TooLargeContent` for relevant 403 status

### LINKED ISSUE #1104: New exception class for too large content
Hi, I got the following exception `Octokit::Forbidden` (masked my repository info) when running [`Octokit.contents`](http://octokit.github.io/octokit.rb/Octokit/Client/Contents.html#contents-instance_method) 👇 

```
Octokit::Forbidden · GET https://api.github.com/repositories/***/contents/***.js?ref=***:
403 - This API returns blobs up to 1 MB in size. The requested blob is too large to fetch via the API,
but you can use the Git Data API to request blobs up to 100 MB in size.
Error summary: resource: Blob field: data code: too_large
// See: https://developer.github.com/v3/repos/contents/#get-contents
```

I want to handle the exception as a different exception as `Octokit::Forbidden`, such as `Octokit::TooLargeContent`, and does this make sense?

What do you think about this idea? I would be happy if there would be any feedback. 🙇 

See also https://developer.github.com/v3/repos/contents/#get-contents

ybiquitous:
I try adding the new exception class `Octokit::TooLargeContent`. Please give me any feedback. 🙏 

Fix #1104

### DIFF
--- a/lib/octokit/error.rb
+++ b/lib/octokit/error.rb
@@ -63,6 +63,8 @@ module Octokit
         Octokit::TooManyRequests
       elsif body =~ /login attempts exceeded/i
         Octokit::TooManyLoginAttempts
+      elsif body =~ /returns blobs up to [0-9]+ MB/i
+        Octokit::TooLargeContent
       elsif body =~ /abuse/i
         Octokit::AbuseDetected
       elsif body =~ /repository access blocked/i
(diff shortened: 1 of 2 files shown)


ybiquitous:
@tarebyte Thanks for your review! Is there something which I should do to merge this PR?

tarebyte:
Nope! Was just waiting for CI to pass 😄

ybiquitous:
Thank you!

PR #1112, opened 2019-03-28
Set up CI with Azure Pipelines [skip ci]

tarebyte:


tarebyte:
Oh lame, so as it turns out there isn't support for the latest ruby versions

```
##[error]Version spec 2.5.5 for architecture %s did not match any version in Agent.ToolsDirectory.
Available versions: /opt/hostedtoolcache
2.3.7,2.4.4,2.5.1
If this is a Microsoft-hosted agent, check that this image supports side-by-side versions of Ruby at https://aka.ms/hosted-agent-software.
If this is a self-hosted agent, see how to configure side-by-side Ruby versions at https://go.microsoft.com/fwlink/?linkid=2005989.
##[section]Finishing: UseRubyVersion
```

Issue #1088, opened 2018-11-12
Initialize ignores falsy values


nickstanish:
https://github.com/octokit/octokit.rb/blob/737748e4e5f2de23e7744b5e9c1ec57133d6ebad/lib/octokit/client.rb#L122

If a value is set to `false` then initialize will fallback to the default configuration for that key. In most cases this shouldn't matter because nothing is set, but it will likely lead to future bugs.

For example, initializing with `{ :auto_paginate => false }` would set `nil` rather than `false`.

tarebyte:
Yeah I think that's a reasonable expectation, mind opening a pull request?

PR #1117, opened 2019-04-21
Overwrite nil Client parameters with defaults

### LINKED ISSUE #1113: octokit 4.14.0 / berkshelf regression - api_endpoint nil url
When using berkshelf `7.0.8` with octokit `4.13.0` and cookbooks hosted on github, it is not required to specify `api_endpoint` configuration in the berkshelf configuration. 
After upgrading octokit to `4.14.0`, berkshelf crashes with a nil url - adding an `api_endpoint` config gets things working again (as does pinning octokit to `4.13.0`). 
I haven't dug any further into this, so I couldn't say whether it's octokit or berkshelf doing something wrong, but from the outside `4.14.0` has a regression compared to `4.13.0`.

akerl:
This should fix #1113 . The issue there affected 2 of my gems as well. It looks like the fix for #1088 accidentally affected `nil` values as well as `false`. Because `nil` is a common pattern for pass-through configuration, parameters set to nil should still be overridden by module defaults.

For example, if a Ruby CLI uses octokit under the hood, and wants to allow configuration of the API endpoint, it might do the following:

```
@client ||= Octokit::Client.new(
  access_token: auth.token,
  auto_paginate: true,
  default_media_type: 'application/vnd.github.moondragon+json',
  api_endpoint: @options[:api_endpoint]
)
```

In v4.14.0, if the user *doesn’t* provide an api_endpoint, and that line resolves to `api_endpoint: nil`, octokit returns an error.

This PR corrects that so that false is still reliably saved as the client setting, but nil is overridden.


mistydemeo:
Thanks for opening this - I was bit by this too!

stanhu:
Ran into this as well.

nickfloyd:
Hey @akerl thank you for the fix/contribution here.  It's a great catch.  Given that this was built against master and that branch has deviated significantly from our current release branch (we'll be working on getting this sorted eventually - sorry for the trouble) I am going to open a new branch with the change you made and a test to back it up and close this PR.

I'd appreciate your 👀  on the change if you have time. I'll make sure the new changeset is linked to this one and I'll also make sure to tag you in the PR just to connect the dots.  Either way, we'll get this change in the next release of v4.

PR #1121, opened 2019-05-10
Fix service status methods. Fixes #1107

### LINKED ISSUE #1107: status.github.com is deprecated
As noted in [this blog post](https://github.blog/2018-12-11-introducing-the-new-github-status-site/) dated 2018-12-11, the old github status site ([status.github.com](https://status.github.com)) is deprecated. It has been replaced with a new site located at [www.githubstatus.com](www.githubstatus.com).

It looks like octokit directly calls out to the old api: https://github.com/octokit/octokit.rb/blob/master/lib/octokit/client/service_status.rb#L11

As of 2019-03-04, the old API will be disabled entirely: https://twitter.com/githubstatus/status/1098992086970454016

octokit should be updated to use the api from the new status page, which is documented [here](https://www.githubstatus.com/api) if this functionality is still desired.

vierarb:
This Pull Request fixes #1107 the service status methods that were using the old [status.github.com](https://status.github.com), replacing them with the updated URL [githubstatus.com](https://www.githubstatus.com).

Also a new method was added and tested: `github_status_summary` that retrieves both the `status` and the `components` (status messages).

One thing to take into account is that with the new API endpoints, the structure has changed, so the old `body` node is now called `description` inside each component, breaking some backwards compatibility.


josacar:
@tarebyte how does this look to you?

nickfloyd:
Marked as breaking (due to new endpoints returning a slightly different body structure, but not significant enough to bump the major version IMO.

PR #1132, opened 2019-07-09
Issue 1126 branch protection accept header

spikex:
This bumps the accept header use for branch protection from `application/vnd.github.loki-preview+json` to `application/vnd.github.luke-cage-preview+json` allowing the `required_approving_review_count` feature to be used. 

It also fixes the test for branch protection, which worked against the VCR cassettes, but not the live API.


spec/octokit/client/repositories_spec.rb
```diff
@@ -390,22 +406,51 @@
...
+          required_status_checks: {
+            strict: true,
+            contexts: []
+          },
+          enforce_admins: true,
+          required_pull_request_reviews: nil,
+          restrictions: nil
+        }
+
+        @client.protect_branch(@repo.full_name, "master",
```
kytrinyx: We probably break long lines all over the place in this codebase, but my personal preference is long lines rather than broken lines, unless it gets excessive and ridiculous. Not something to change here, I'm just mentioning it in case everyone is like "oh, yes thank goodness, let's do that going forward".

spec/octokit/client/repositories_spec.rb
```diff
@@ -381,7 +388,16 @@
...
-        branch = @client.protect_branch(@repo.full_name, "master", accept: preview_header)
+        branch = @client.protect_branch(@repo.full_name, "master",
+                                        { accept: preview_header,
+                                          required_status_checks: {
+                                            strict: true,
+                                            contexts: []
+                                          },
+                                          enforce_admins: true,
+                                          required_pull_request_reviews: nil,
+                                        }
```
tarebyte: Could we reuse the `protect_params` hash created above? It looks like they are the same thing.

spec/octokit/client/repositories_spec.rb
spikex: Got it, I've unwrapped it.

spikex:
Thanks! After the fact I also though that the VCR change probably should have been it's own thing.

spec/octokit/client/repositories_spec.rb
spikex: That last test is in a different block for some reason. Probably worth reorganizing at some point.

PR #1133, opened 2019-07-12
Add support for Template Repositories API

EricPickup:
[Template Repositories](https://github.blog/2019-06-06-generate-new-repositories-with-repository-templates/) is a feature shipped last month that allows you to designate a repository as a template, and allows others to create a clone of that template repository.

The preview (not sure if I should link here since it's internal) was recently released for our REST api. 

This PR adds the ability to do the following with Octokit:
- Use a template repository to generate a new repository
- Viewing whether a repository is a template
- Creating a template repository
- Updating a repository to change whether or not it's a template

(closes https://github.com/education/classroom/issues/1941)


lib/octokit/client/repositories.rb
```diff
@@ -155,6 +156,7 @@ def create_repository(name, options = {})
         opts = options.dup
         organization = opts.delete :organization
         opts.merge! :name => name
+        ensure_api_media_type(:template_repositories, opts) if opts.include? :is_template
```
EricPickup: This is where I'm at a crossroads.
Adding support for the preview in our existing method `create_repository` vs. isolating it into its own method (e.g. `create_template_repository`).
You can see I did this for `template_repository` - I could've easily implemented it in the main `edit_repository` method like I did here with `create_repository`.
What do you think @tarebyte?

lib/octokit/client/repositories.rb
```diff
@@ -696,6 +715,30 @@ def update_subscription(repo, options = {})
...
+      # Check if a repository is a template repository
+      #
+      # @param repo [Integer, String, Hash, Repository] A GitHub repository
+      # @return [Boolean] True if repository is a template repository, false otherwise.
+      def template_repository?(repo)
+        options = ensure_api_media_type(:template_repositories, {})
+        response = get Repository.path(repo), options
+        response.is_template?
+      end
+      alias :template_repo? :template_repository?
```
tarebyte: I'm kind of against a helper method like this being in the client.

The only time we support `def *?` methods is when we are expecting a `204` response.

https://github.com/octokit/octokit.rb/blob/9bc8cd25c335cadbc898e11976a32aa8ce77ec8c/lib/octokit/client/organizations.rb#L193-L211

lib/octokit/client/repositories.rb
```diff
@@ -696,6 +715,30 @@ def update_subscription(repo, options = {})
...
+      # @param repo [String, Hash, Repository] A GitHub repository
+      # @param is_template [Boolean] True if making repo a template, false if making it not a template.
+      # @return [Sawyer::Resource] Repository information
+      def template_repository(repo, is_template)
+        repo = Repository.new(repo)
+        options = { is_template: is_template, name: repo.name }
+        options = ensure_api_media_type(:template_repositories, options)
+        patch "repos/#{repo}", options
+      end
+      alias :template_repo :template_repository
```
tarebyte: Following up on your comment https://github.com/octokit/octokit.rb/pull/1133/files#r303087310

I would rather us update the `edit_repository` method than create a new helper method for this. It fits the pattern of the client better.

lib/octokit/client/repositories.rb
```diff
@@ -192,6 +194,23 @@ def transfer_repository(repo, new_owner, options = {})
...
+      # @option options [String] :private `true` makes the repository private, and `false` makes it public.
+      # @return [Sawyer::Resource] Repository info for the new repository
+      def clone_template_repository(repo, name, options = {})
+        options.merge! :name => name
+        options = ensure_api_media_type(:template_repositories, options)
+        post "#{Repository.path repo}/generate", options
+      end
+      alias :create_repository_from_template :clone_template_repository
+      alias :clone_template_repo :clone_template_repository
+      alias :create_repo_from_template :clone_template_repository
```
tarebyte: There are a lot of aliases here, could we just keep `clone_template_repo` ?

lib/octokit/client/repositories.rb
tarebyte: I'm 👍 for updating the existing methods `create_repository` and `edit_repository` rather than creating a new method just for this option.

It keeps things more consistent and I think it would be an unexpected side effect if these methods didn't support these options.

lib/octokit/client/repositories.rb
EricPickup: Makes sense. I added this to avoid updating the existing `get repository` method with the preview key.

lib/octokit/client/repositories.rb
EricPickup: Done, I've actually changed the name completely to `create_github_repository_from_template` with one alias: `create_github_repo_from_template`

EricPickup:
@tarebyte looks like CI failed from a flaky build. Would you mind restarting it?

lib/octokit/client/repositories.rb
```diff
@@ -26,6 +26,7 @@ def repository?(repo, options = {})
       # @param repo [Integer, String, Hash, Repository] A GitHub repository
       # @return [Sawyer::Resource] Repository information
       def repository(repo, options = {})
+        options = ensure_api_media_type(:template_repositories, options)
```
tarebyte: The more I think about it, I'd rather leave out the preview here, if someone wants the new data they can opt into it.

Mind if we remove this line?

lib/octokit/client/repositories.rb
```diff
@@ -41,11 +42,13 @@ def repository(repo, options = {})
       # @option options [String] :private `true` makes the repository private, and `false` makes it public.
       # @option options [String] :has_issues `true` enables issues for this repo, `false` disables issues.
       # @option options [String] :has_wiki `true` enables wiki for this repo, `false` disables wiki.
+      # @option options [Boolean] :is_template `true` makes the repository a template, `false` makes it not a template.
       # @option options [String] :has_downloads `true` enables downloads for this repo, `false` disables downloads.
       # @option options [String] :default_branch Update the default branch for this repository.
       # @return [Sawyer::Resource] Repository information
       def edit_repository(repo, options = {})
         repo = Repository.new(repo)
+        options = ensure_api_media_type(:template_repositories, options) if options.include? :is_template
```
tarebyte: I'm usually a fan of guard clauses, but this feels a bit terse.

Can we update it?

```ruby
if options.include? :is_template
  options = ensure_api_media_type(:template_repositories, options)
end
```

lib/octokit/client/repositories.rb
```diff
@@ -155,6 +159,7 @@ def create_repository(name, options = {})
         opts = options.dup
         organization = opts.delete :organization
         opts.merge! :name => name
+        opts = ensure_api_media_type(:template_repositories, opts) if opts.include? :is_template
```
tarebyte: Same query here about making it multiline

lib/octokit/client/repositories.rb
EricPickup: Do you mean replace it with a check like the others methods?

EricPickup:
@tarebyte looks like CI failed from a flaky test

PR #1134, opened 2019-07-16
Fix example in #update_branch documentation to call the correct method

rmacklin:
The example in the documentation for `update_branch` was calling `update_ref`:

image

This PR corrects the example to call `update_branch` instead.


tarebyte:
🍰

PR #1135, opened 2019-07-17
Update reference to "add specs" step in CONTRIBUTING

gallexi:
Hello! 

I was reading through the contributing docs, and noticed a bit that seems to want us to go back and write more tests (step 3), but actually says to go back to step 4, which is to run the tests.

Thanks!


tarebyte:
Thanks @gallexi ✨

PR #1137, opened 2019-07-17
Clarify force vs fast-forward updates in #update_ref and #update_branch docs

rmacklin:
Previously the description of the `force` parameter was:
> A flag indicating one wants to force the update to make sure the update is a fast-forward update.

which is contradictory: a force update does *not* make sure the update is a fast-forward update -- rather, the opposite.

The docs also link to the official API docs for the update-a-reference endpoint:
https://developer.github.com/v3/git/refs/#update-a-reference

There, the description of the force parameter is:
> Indicates whether to force the update or to make sure the update is a fast-forward update

It's almost the same, but has the crucial "or" which was missing in the octokit documentation. So here, I've updated the octokit documentation to mirror the wording on the official API docs. Additionally, I've added an example of how to do a fast-forward update with `force` set to `false`.

### DIFF
--- a/lib/octokit/client/refs.rb
+++ b/lib/octokit/client/refs.rb
@@ -65,6 +65,8 @@ module Octokit
       # @see https://developer.github.com/v3/git/refs/#update-a-reference
       # @example Force update heads/sc/featureA for octocat/Hello-World with sha aa218f56b14c9653891f9e74264a383fa43fefbd
       #   Octokit.update_ref("octocat/Hello-World", "heads/sc/featureA", "aa218f56b14c9653891f9e74264a383fa43fefbd")
+      # @example Fast-forward update heads/sc/featureA for octocat/Hello-World with sha aa218f56b14c9653891f9e74264a383fa43fefbd
+      #   Octokit.update_ref("octocat/Hello-World", "heads/sc/featureA", "aa218f56b14c9653891f9e74264a383fa43fefbd", false)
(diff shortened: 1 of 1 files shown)


rmacklin:
@tarebyte Do you have any thoughts on this?

onen954:
> Previously the description of the `force` parameter was:
> 
> > A flag indicating one wants to force the update to make sure the update is a fast-forward update.
> 
> which is contradictory: a force update does _not_ make sure the update is a fast-forward update -- rather, the opposite.
> 
> The docs also link to the official API docs for the update-a-reference endpoint:
> https://developer.github.com/v3/git/refs/#update-a-reference
> 
> There, the description of the force parameter is:
> 
> > Indicates whether to force the update or to make sure the update is a fast-forward update
> 
> It's almost the same, but has the crucial "or" which was missing in the octokit documentation. So here, I've updated the octokit documentation to mirror the wording on the official API docs. Additionally, I've added an example of how to do a fast-forward update with `force` set to open

rmacklin:
Hi @hmharvey, thanks for taking a look at this PR!

I see that you've merged the `4-stable` branch into this PR. If you'd rather I rebase this branch to get rid of the merge commits, I'd be happy to do so. Let me know!

hmharvey:
Sure if you don't mind @rmacklin! Thank you ✨

rmacklin:
Done!

PR #1149, opened 2019-09-16
Generate deployment client from the OpenAPI spec

hmharvey:
Part of https://github.com/octokit/octokit.rb/issues/1157. An updated spike of https://github.com/octokit/octokit.rb/pull/1010 as the underlying api spec has changed. So much thanks to `@kytrinyx` for her work and `@tarebyte` for helping me get started ✨ 

This includes a breaking change for `deployment_statuses` and `create_deployment_status`. We're updating the parameter from `deployment_url` to use both `repo` and `deployment_id` as this is more consistent with other methods.

This also includes a new method, `deployment_status`, that was previously unsupported.

TODO: 
- [x] Update testing for the breaking change in the client
- [x] Update the name from Spike


lib/spike.rb
```diff
@@ -0,0 +1,257 @@
...
+    end
+
+    def method_implementation
+      [
+        *option_overrides,
+        api_call,
+      ].reject(&:empty?).join("\n        ")
+    end
+
+    # TODO. possible breaking change as the url param isn't a required param
```
hmharvey: Previously, the `status` methods (i.e.  for `get-a-single-deployment-status`, `list-deployment-statuses`) had a required [`deployment_url` parameter](https://github.com/octokit/octokit.rb/blob/a1eeda95fa5fcbb532c411e3851e2d8ab894a5ea/lib/octokit/client/deployments.rb#L51). However, the [current OpenAPI spec](https://github.com/octokit/routes/blob/master/openapi/api.github.com/operations/repos/list-deployment-statuses.json#L12) doesn't have this as a required parameter and has `repo` and `deployment_id` instead. 

Going forward, do we want to focus on supporting the existing format and ignore the new parameters so it's not a breaking change?

lib/spike.rb
```diff
@@ -213,120 +102,126 @@ def option_overrides
...
     def alternate_name
-      return unless definition.verb == "GET"
-      return if resource.singular?
-      "list_#{resource.name}"
+      return unless verb == "GET"
+      return if singular?
+      "list_#{namespace}"
     end
 
+    # TODO. `priority` is calculating the wrong priorities bc of this method
```
hmharvey: I would appreciate some 👀 here as I'm having trouble figuring out what this should be calculating for the next method `priority` to work correctly

lib/octokit/client/deployments.rb
```diff
@@ -4,20 +4,11 @@ class Client
     #
     # @see https://developer.github.com/v3/repos/deployments/
     module Deployments
-      # Get a single deployment
```
hmharvey: Still working on the order these methods show up

lib/octokit/client/deployments.rb
```diff
@@ -33,58 +24,71 @@ def deployments(repo, options = {})
...
 
       # List deployment statuses
       #
-      # @param deployment_url [String] A URL for a deployment resource.
-      # @return [Array<Sawyer::Resource>] A list of deployment statuses
+      # @param repo [Integer, String, Repository, Hash] A GitHub repository
+      # @param deployment_id [Integer] The ID of the deployment
+      # @return <Sawyer::Resource> A single deployment statuses
       # @see https://developer.github.com/v3/repos/deployments/#list-deployment-statuses
-      def deployment_statuses(deployment_url, options = {})
```
hmharvey: This potential change is open for discussion below

lib/spike.rb
hmharvey: Making it a breaking change 🎉

lib/openapi_client_generator.rb
```diff
@@ -0,0 +1,241 @@
...
+          path.endpoints.each do |endpoint|
+            arr << OpenAPIClientGenerator::Endpoint.new(endpoint, parameterizer: parameterizer)
+          end
+        end
+        yield new(resource, endpoints: endpoints)
+      end
+    end
+
+    def self.resource_for_path(path)
+      return :unsupported unless path.include? "deployment"
```
tarebyte: Can we make this into an Array? So that we can just add more resources easier?

lib/openapi_client_generator.rb
```diff
@@ -0,0 +1,241 @@
...
+      path_segments = path.split("/").reject{ |segment| segment == "" }
+      resource = path_segments[3]
```
tarebyte: This seems pretty magic, does this ever change?

hmharvey: It definitely will, but it's pretty consistent for the paths under "repos" which was my next focus. I can look further into it, but I felt like it was good enough for now 🤔

lib/openapi_client_generator.rb
```diff
@@ -0,0 +1,241 @@
+require 'forwardable'
```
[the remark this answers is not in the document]
hmharvey: doesn't look like it did 🎉

lib/openapi_client_generator.rb
```diff
@@ -206,10 +212,11 @@ def self.at(definition, parameterizer: OpenAPIClientGenerator::Endpoint::Positio
     end
 
     def self.resource_for_path(path)
-      return :unsupported unless path.include? "deployment"
       path_segments = path.split("/").reject{ |segment| segment == "" }
       resource = path_segments[3]
-      resource ||= "repositories"
+
+      supported_resources = ["deployments"]
+      return (supported_resources.include? resource) ? resource : :unsupported
```
hmharvey: @tarebyte how does this look? I don't believe `return :unsupported unless path.include?(["deployment"])` is valid 🤔

hmharvey:
> This looks ✨ I don't want to merge it yet because we need to figure out how to keep a `4.x` branch and a new `5.x` branch instead of just merging master while we develop this.

Definitely 👍

PR #1151, opened 2019-09-18
adds support for update_pull_request_review in reviews.rb

eric-silverman:
Adds support for [Update Pull Request Review](https://developer.github.com/v3/pulls/reviews/#update-a-pull-request-review) in reviews.rb


lib/octokit/client/reviews.rb
```diff
@@ -204,6 +204,24 @@ def delete_pull_request_review_request(repo, id, reviewers={}, options = {})
         options = options.merge(reviewers)
         delete "#{Repository.path repo}/pulls/#{id}/requested_reviewers", options
       end
+
+
```
tarebyte:
```suggestion
```

lib/octokit/client/reviews.rb
```diff
@@ -204,6 +204,24 @@ def delete_pull_request_review_request(repo, id, reviewers={}, options = {})
...
+      # @see https://developer.github.com/v3/pulls/reviews/#update-a-pull-request-review
+      #
+      # @example
+      #   @client.update_pull_request_review('octokit/octokit.rb', 825, 6505518, 'This is close to perfect! Please address the suggested inline change. And add more about this.')
+      #
+      # @return [Sawyer::Resource] Hash representing the review comment
+      def update_pull_request_review(repo, number, review, body)
+        put "#{Repository.path repo}/pulls/#{number}/reviews/#{review}", {body: body}
+      end
+
```
tarebyte:
```suggestion
```

spec/octokit/client/reviews_spec.rb
```diff
@@ -58,6 +58,17 @@
     end
   end
 
+  describe '.update_pull_request_review' do
```
tarebyte: Instead of hardcoding this, can we add a new context "with pull request review" and move this under there?

lib/octokit/client/reviews.rb
```diff
@@ -204,6 +204,24 @@ def delete_pull_request_review_request(repo, id, reviewers={}, options = {})
...
+      # @param body [String] body text of the pull request review.
+      # @see https://developer.github.com/v3/pulls/reviews/#update-a-pull-request-review
+      #
+      # @example
+      #   @client.update_pull_request_review('octokit/octokit.rb', 825, 6505518, 'This is close to perfect! Please address the suggested inline change. And add more about this.')
+      #
+      # @return [Sawyer::Resource] Hash representing the review comment
+      def update_pull_request_review(repo, number, review, body)
+        put "#{Repository.path repo}/pulls/#{number}/reviews/#{review}", {body: body}
+      end
```
tarebyte: This is missing the `options` hash which allows us to pass in things like preview headers.

WDYT about something like this:

```ruby
def update_pull_request_review(repo, number, review, body, options = {})
  options[:body] = body
  put "#{Repository.path repo}/pulls/#{number}/reviews/#{review}", options
end
```

spec/octokit/client/reviews_spec.rb
eric-silverman: works for me, updated

lib/octokit/client/reviews.rb
eric-silverman: works for me, updated

eric-silverman:
@tarebyte should be good now

PR #1153, opened 2019-09-24
Add `delete_installation` method to uninstall a GitHub App

yykamei:
There's an API endpoint to let GitHub Apps uninstall themselves from users' installation. This PR adds this feature to Octokit.

https://developer.github.com/v3/apps/#delete-an-installation


lib/octokit/client/apps.rb
```diff
@@ -204,6 +204,19 @@ def find_installation_repositories_for_user(installation, options = {})
...
+      # Delete an installation and uninstall a GitHub App
+      #
+      # @param installation [Integer] The id of a GitHub App Installation
+      # @param options [Hash] A customizable set of options
+      #
+      # @see https://developer.github.com/v3/apps/#delete-an-installation
+      #
+      # @return [Boolean] Success
+      def delete_installation(installation, options = {})
+        opts = ensure_api_media_type(:uninstall_github_app, options)
```
yykamei: In [this](https://developer.github.com/v3/apps/#delete-an-installation) document, this endpoint seems to require two values for the `Accept` header, but I succeeded to uninstall my actual installation on GitHub only with the media type `application/vnd.github.gambit-preview+json`. So, I set only `uninstall_github_app`.

PR #1154, opened 2019-09-27
Update to inherit from correct Faraday error

Gasparila:
The update to Faraday is causing `require "octokit"` to break. Updating
to `Faraday::ClientError` fixes it and is correct based on Faraday code:
https://github.com/lostisland/faraday/blob/master/lib/faraday/error.rb

Before:
```
script/test
...
/Users/kevinmiller/octokit.rb/lib/octokit/middleware/follow_redirects.rb:14:in `<module:Middleware>': uninitialized constant Faraday::Error::ClientError (NameError)
Did you mean?  Faraday::ClientError
```

After:
```
script/test
...
Finished in 26.83 seconds (files took 0.79837 seconds to load)
775 examples, 0 failures
```


Gasparila:
Looks like the failed test is flaky given that it is to do with caching and unrelated to my change? 🤞

BanzaiMan:
This will then assume that Faraday version is 0.16.0 and up. Some sort of fallback is more desirable.

ushis:
@gas You could add faraday as dependency to the gemspec with `>= 0.16.0`

michaelherold:
> This will then assume that Faraday version is 0.16.0 and up. Some sort of fallback is more desirable.

[v0.15.0](https://github.com/lostisland/faraday/blob/v0.15.0/lib/faraday/error.rb) uses this organization but has an alias that allows it to still work.

This organization has been the current one since at least 0.9.0 (which was a long-standing release). They just removed the backwards-compatibility with 0.16.0.

I suggest this change be merged.

michaelherold:
Specifically, it's [this commit](https://github.com/lostisland/faraday/commit/793e6ba9c94ad06fd458b3a2c10bce77abe6917e#diff-1b35e95f85a819bcd3c16d920ecdcffe) where they remove the aliases (see [here](https://github.com/lostisland/faraday/commit/793e6ba9c94ad06fd458b3a2c10bce77abe6917e#diff-1b35e95f85a819bcd3c16d920ecdcffeL79-L81) for the specific lines).

avondohren:
If the change was present in Faraday v0.9.0, then we would only need to update the gemspec with `>= 0.9.0`, correct?

Gasparila:
@avondohren I agree, currently  "> 0.8" is required. Added `>= 0.9` explicit requirement in `11c65a6b`

michaelherold:
I submitted a patch to restore the backward-compatibility layer to Faraday's 0.16.x releases. I think they want to drop it for 1.0 so this change should still be merged into Octokit.

octokit.gemspec
```diff
@@ -6,6 +6,7 @@ require 'octokit/version'
 Gem::Specification.new do |spec|
   spec.add_development_dependency 'bundler', '>= 1', '< 3'
   spec.add_dependency 'sawyer', '>= 0.5.3', '~> 0.8.0'
+  spec.add_dependency 'faraday', '>= 0.9', '< 2.0'
```
michaelherold: This will pick up v0.16.0, which is the affected version. I think we just want to change the constant.

```suggestion
  spec.add_dependency 'faraday', '>= 0.9'
```

If we want a version constraint, we want it as permissive as possible.

tarebyte:
@Gasparila thanks for the PR, am I right in thinking that requiring Faraday `>= 9.0` that we'll break the gem for users who are using Sawyer versions that are older than `0.8.2`?

At least that's how I understand what I'm reading on https://rubygems.org/gems/sawyer/versions/0.8.1

Gasparila:
@tarebyte it looks like Sawyer supported 0.9 since version 0.1 (https://rubygems.org/gems/sawyer/versions/0.1.0)

The `~> 0.8` is shorthand for `>= 0.8 < 1.0` (the last digit can fluctuate)

tarebyte:
> The `~> 0.8` is shorthand for `>= 0.8, < 1.0` (the last digit can fluctuate)

TIL I have no idea how Rubygems short hand works.

If you wouldn't mind accepting https://github.com/octokit/octokit.rb/pull/1154#discussion_r329212579 I think this'll be good to merge.

Thanks ✨

Gasparila:
> If you wouldn't mind accepting #1154 (comment) I think this'll be good to merge

Hmmm I disagree that we should be permissive of major version bumps, and all of the other dependencies specify a cap of the next major version. Originally, I had `~> 0.9`, but that would disallow `1.x` releases and this seems more permissive. Happy to make the change, but that would make faraday handled differently from the other dependencies (`sawyer` and `bundler`)

BobbyMcWho:
@Gasparila IMO you shouldn't restrict versions on the upperside unless you know for sure that a new version of the dependency is going to break something that this package uses. If you leave the top open, then you allow folks using this along with other gems that have the same dependency to keep upgrading. Once a breaking change is released, then you cap the upper bound if you need to as a patch release, or release a compatibility fix as appropriate.

Gasparila:
> Once a breaking change is released, then you cap the upper bound if you need to as a patch release, or release a compatibility fix as appropriate.

Or you don't notice it and then your gem is broken, necessitating another PR like this one 😉 

Regardless, I care most about getting this fixed so updated in d4d77dc

michaelherold:
> Or you don't notice it and then your gem is broken, necessitating another PR like this one

In this case, the Faraday maintainers leaned into the "less than 1.0 means all bets are off" clause of SemVer. :grin:

Because Faraday is such a common "plumbing" gem that ends up being a transitive dependency across libraries, I've found that keeping it with a relaxed constraint leads to a better user experience -- even when they break API compatibility -- because it avoids dependency hell where you have to fork it to unblock unrelated gem updates.

acidprime:
Any update on this PR, whats needed or what it needs to be merged , I'm using [slack-ruby-client](https://github.com/slack-ruby/slack-ruby-client)  and this gem which is in a dependency loop until this is solved by one or all of the gems. Really don't want to fork my production deployment if possible.

BobbyMcWho:
@acidprime try a `bundle update faraday`, version 0.16.2 should have resolved most downstream issues

acidprime:
> @acidprime try a `bundle update faraday`, version 0.16.2 should have resolved most downstream issues

Interesting, I tried pinning the following and I'm still getting the error

```ruby
ruby '2.5.1'

source 'https://rubygems.org'

gem 'async-websocket', '~>0.8.0'
gem 'aws-sdk', '3.0.1'
gem 'faraday','0.16.2'
gem 'octokit', '4.14.0'
gem 'rugged', '0.27.5'
gem 'sawyer', '0.8.0'
gem 'slack-ruby-bot', :git => 'https://github.com/slack-ruby/slack-ruby-bot.git'

```

BobbyMcWho:
And you're 100% sure you ran `bundle update faraday` after pinning it?

acidprime:
> And you're 100% sure you ran `bundle update faraday` after pinning it?

Its all ran in docker with no cache, so there is no "update" that would be relevant I believe.

PR #1158, opened 2019-10-06
add create_branch method to refs.rb

alon-k:
refs.rb already contains helper methods made specifically for updating
and deleting branches. This adds a helper method for creating a branch
as well.


nickfloyd:
Hey @alon-k I know it's been a while 😬 . Thank you for this change ❤️ .  Because we have not maintained this for a while and our current working branch (`4-stable`) has really deviated from what's in master. 

I'd like to add this change but it really needs to be added using a branch/fork from the latest release.  I am going to close this PR and add a label to it so that when we have time (or if you do) we can migrate this change to `4-stable`

Issue #1157, opened 2019-10-03
[Tracking issue] Generating client code from the OpenAPI schema


indigok:
Opening this to track the progress of generating the client code from the OpenAPI spec in the shared [routes](https://github.com/octokit/routes) directory.

:eyes: https://github.com/octokit/octokit.rb/pull/1010 for proof of concept
:eyes: https://github.com/octokit/routes/pull/168 for more context

zeke:
Wow! So cool to see this happening.

cc @sarahs @jleaver

hmharvey:
With the current naming schema, we’ve generated 33/91 clients. Thus, as we’ve reached a third of the way through the project and finally have several moving parts solidified, I’m going to take a moment to write some documentation. 

#### Steps to generate a new client

0. Have both [`octokit/routes`](https://github.com/octokit/routes) and `octokit/octokit.rb` in the same directory. Everything else will be run from inside of `octokit.rb`
1. Add the client you want to generate to the [supported resources](https://github.com/octokit/octokit.rb/blob/master/lib/openapi_client_generator.rb#L408). The client names mirror the documentation url after v3, i.e. pull request comments go from `https://developer.github.com/v3/pulls/comments/` to `pulls_comments`.
2. Run `bundle exec rake generate ` to generate all the clients listed in the above line from the openapi spec. 
3. Compare the diff and see if the generated client has any mistakes. The old clients will likely be in a different file but try to find all the corresponding methods - checking the documentation link is a good way to ensure you're finding the right ones. The clients don’t have to be an exact match but should be understandable and useable.
4. Change anything you need to in `openapi_client_generator`.  A high risk area is the `namespace` method. As the API becomes more standardized, hopefully we can reassess this method but as it stands, it’s one of the hardest parts to generate and there’s a lot of special casing. `binding.pry` is your friend.
5. Regenerate and tweak the generator as needed until you feel comfortable pushing and merging it 🚀  

#### Tests
The new way of testing is outlined in #1258. Ideally, if you’re working on a client that contains one of these scenarios, go ahead and add the matching test.

#### Openapi schema
This generator is currently being pulled from `octokit/routes` which scrapes the developer docs. In the near future, the new source of truth will be [`openapi`](https://unpkg.com/browse/@github/openapi@2.6.0/) and we’ll need to make the switch. The main change appears to be switching from a JSON to a YAML schema, but most of the fields are the same.

[...]

PR #1159, opened 2019-10-08
Generate pages client

hmharvey:
Part of #1157 (and built on top of #1149). This generalizes the `openapi_client_generator` to generate the Pages client as well. This includes support for **three** new endpoints: [`update_pages_site`](https://developer.github.com/v3/repos/pages/#update-information-about-a-pages-site),[`enable_pages_site`](https://developer.github.com/v3/repos/pages/#enable-a-pages-site), and [`disable_pages_site`](https://developer.github.com/v3/repos/pages/#disable-a-pages-site). 🎉 

The only breaking change is we decided to remove the alias'd definitions - which retroactively affects the deployments client as well.

For a more accurate diff of files changed within the Pages PR and not the previous PRs it's built on, look [here](https://github.com/hmharvey/octokit.rb/compare/deployment-spike...generate-pages-client).

TODO:
- [x] fix definitions and testing for the two new methods. 
  - Hoping to update the docs for `update pages` as it requires one of multiple parameters which isn't documented but it won't be incorporated into OpenAPI for now
- [x] looks like general testing for pages with CI as well


lib/octokit/client/pages.rb
```diff
@@ -1,63 +1,89 @@
...
-      # @see https://developer.github.com/v3/repos/pages/#list-a-specific-pages-build
-      # @example
-      #   Octokit.pages_build("github/developer.github.com", 5472601)
-      def pages_build(repo, id, options = {})
-        opts = ensure_api_media_type(:pages, options)
-        get "#{Repository.path repo}/pages/builds/#{id}", opts
+      # @param build_id [Integer] The ID of the build
+      # @return <Sawyer::Resource> A single pages build
+      # @see https://developer.github.com/v3/repos/pages/#get-a-specific-pages-build
+      def pages_build(repo, build_id, options = {})
```
tarebyte: Naming seems a little odd, not sure if it's possible to make it more clear through auto generation.

spec/octokit/client/pages_spec.rb
```diff
@@ -5,56 +5,93 @@
   before do
     Octokit.reset!
     @client = oauth_client
+    @pages_repo =  "#{test_github_login}/#{test_github_login}.github.io"
```
tarebyte: Could we make/ensure this repo exists before the tests start? Maybe in an RSpec context?

spec/octokit/client/pages_spec.rb
```diff
@@ -5,56 +5,93 @@
...
 
-  describe ".latest_pages_build", :vcr do
-    it "lists information about the latest page build" do
-      latest_build = @client.latest_pages_build("github/developer.github.com")
-      expect(latest_build.status).to eq("built")
-      assert_requested :get, github_url("/repos/github/developer.github.com/pages/builds/latest")
-    end
-  end # .latest_pages_build
+  context "with build", :vcr do
+      before do
```
tarebyte:
```suggestion
    before do
```

mattlangan:
@hmharvey Is the `enable_pages_site` method supposed to be available in the Octokit Ruby wrapper? I see that method available in this PR, but don't see it in the Pages methods on the current release. I'm not sure if I'm losing history between your fork and the original repo or if I'm just not looking at this correctly.

stevevancleemput:
Well I saw that name passing by when I was browsing in my files ...

hmharvey:
👋🏻  @mattlangan this is a separate project and has no bearing on the `4-stable` branch and the current releases.

PR #1161, opened 2019-10-31
Generate hooks client

hmharvey:
Part of #1157. Continues to generalize the `openapi_client_generator` and generates the `hooks.rb` client. 

There are _breaking_ changes as a few of the method definitions have changed, however the underlying api calls appeared to have changed and the old methods don't completely work. This does removes support for service hooks (which aren't supported in general anymore) and `parse_payload` which is a helper function. Also, `remove_hook` is now `delete_hook` and `test_hook` is `test_push_hook` as these more closely resemble the API.

This also retroactively affects the previous generated clients as we changed list calls to `paginate`.

TODOs:
- [x] add a check that the required key of the config parameter, url, is present
- [x] clean up testing for `hooks_spec` and `openapi_client_generator_spec`


lib/octokit/client/hooks.rb
```diff
@@ -1,296 +1,159 @@
...
-      #
-      # Requires authenticated client.
+      # Delete a hook
       #
-      # @param repo [Integer, String, Hash, Repository] A GitHub repository.
-      # @param id [Integer] Id of the hook to remove.
-      # @return [Boolean] True if hook removed, false otherwise.
+      # @param repo [Integer, String, Repository, Hash] A GitHub repository
+      # @param hook_id [Integer] The ID of the hook
+      # @return <Boolean> True on success, false otherwise
```
tarebyte: Looks like this should be 

```suggestion
      # @return [Boolean] True on success, false otherwise
```

spec/octokit/client/hooks_spec.rb
```diff
@@ -7,16 +7,9 @@
...
-    it "returns all the hooks supported by GitHub with their parameters" do
-      hooks = @client.available_hooks
-      expect(hooks.first.name).to eq("activecollab")
-    end
-  end
-
   context "with repository" do
     before(:each) do
-      @repo = @client.create_repository("an-repo")
+      @repo = @client.create_repository("a-repo")
```
tarebyte: Do we maybe want to use the already created repo?

spec/octokit/client/hooks_spec.rb
```diff
@@ -36,17 +29,20 @@
...
-        @client.remove_hook(@repo.full_name, @hook.id)
+        @client.delete_hook(@repo.full_name, @hook.id)
       end
 
       describe ".create_hook", :vcr do
         it "creates a hook" do
           assert_requested :post, github_url("/repos/#{@repo.full_name}/hooks")
         end
+        it "returns with no config url passed" do
+          expect { @client.create_hook(@repo.full_name)}.to raise_error Octokit::MissingKey
```
tarebyte: ✨

Issue #517, opened 2014-10-07
Octokit.last_response is not updated when calling Octokit.last_response.rels[:next].get


adamjmurray:
Based on the documentation for this library, I thought I could do this:

``` ruby
pull_requests = Octokit.pull_requests repo, state: 'all'
puts pull_requests.first.id
pull_requests = Octokit.last_response.rels[:next].get.data
puts pull_requests.first.id

# last_response is never updated, so attempting to paginate further returns the same set of pull requests as the previous request
pull_requests = Octokit.last_response.rels[:next].get.data
puts pull_requests.first.id
```

But as mentioned in the comment of that code, calling Octokit.last_response.rels[:next].get multiple times keeps returning the same data. I'm guessing the .get() call in that expression is not updating last_response. Since I am making an API request, I would expect last_response to be updated and this code to work.

As a workaround, it looks like I can do this:

``` ruby
pull_requests = Octokit.pull_requests repo, state: 'all'
puts pull_requests.first.id

last_response = Octokit.last_response.rels[:next].get
pull_requests = last_response.data
puts pull_requests.first.id

last_response = last_response.rels[:next].get
pull_requests = last_response.data
puts pull_requests.first.id
```

This works as expected but feels like unnecessary boilerplate code just to do some basic pagination.

I hope this can be improved in a future version of Octokit. Or at least update the documentation to explain this better.

danielbeardsley:
Ran into this same issue and had the same frustrations. The docs are unhelpfully vague regarding proper usage of `last_response` (or even which thing has that method... i.e. is it `Octokit.last_response` or `Client#last_response`?)

I ended up doing the same thing as @adamjmurray mentioned in his workaround.

rileytg:
@pengwynn i'm still getting this issue described in OP. I am working with the commits api so auto pagination won't work... for now I'm going to use the workarounds described here but i think the last_response should be fixed (maybe i can do...) or the doc should be updated (i for sure can do). 

Please advise on what my next steps should be (try to fix last_response, or update docs).

pengwynn:
@rileytg I haven't thought through a robust fix yet. I think it will involve pulling a slimmed down version of the Sawyer dependency into Octokit and wire up the plumbing the way we want it to work.

sschuberth:
> I haven't thought through a robust fix yet.

How come the issue is closed if you admit it's not fixed yet?

PR #1163, opened 2019-11-05
Update Pagination section of README to use client.get

BenMusch:
Based on https://github.com/octokit/octokit.rb/issues/517, the docs are somewhat misleading because the pagination method only works for 1 additional page. Since that's a somewhat rare use-case, it would save some readers confusion to not have to face this bug

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,7 +290,7 @@ link relations](#hypermedia-agent).
 
 ```ruby
 issues = client.issues 'rails/rails'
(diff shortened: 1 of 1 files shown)


BenMusch:
BTW my assumption here is that this readme update will propgate to the docs. If that's not accurate, feel free to link me to where this section of the docs is managed and I can make the change there as well

PR #1173, opened 2019-12-19
Cut a new release

ys:
It has been 9 months and a lot of changes since the last release.

Would be dope to cut a new release here.

Thanks.


tarebyte:
Done, see https://github.com/octokit/octokit.rb/issues/1177#issuecomment-570311185

PR #1178, opened 2020-01-02
Update follow_redirects.rb

user-tony:
double colon


tarebyte:
Thanks for the PR @user-tony, what exactly do we gain by making this change?

akerl:
I suspect this might fix the error I’m seeing:

```/usr/lib/ruby/gems/2.7.0/gems/octokit-4.7.0/lib/octokit/middleware/follow_redirects.rb:14:in `<module:Middleware>': uninitialized constant Faraday::Error::ClientError (NameError)```

zhtw1234:
幹你娘

PR #1183, opened 2020-01-19
enable issue events API to return project events

gammons:
Hi there,

This change allows the `issue_events` api to return project card events, by using the `application/vnd.github.starfox-preview+json` header.  

This change works, but it's currently lacking specs.  I'm thinking the best way to exercise these specs is if one of the admins of this repo could create a project board and put a couple of issues on it.  I can then  record the api response and write specs that exercise the response from there.

But, I'm open to any ideas on how to move forward with specs!


hmharvey:
Hi @gammons 👋 

Specs are created off of test repos as set by [environment variables](https://github.com/octokit/octokit.rb#running-and-writing-new-tests). For this change though, we don't need to rerecord them as the response hasn't really changed, it just allows more events. I've gone ahead and added the preview to the test and so it should be good to merge when the tests pass ✨

PR #1187, opened 2020-01-24
Add the 'branches for commit-sha HEAD' endpoint

tgmachina:
One more preview endpoint, this time to retrieve a set of branches given a commit-sha.

I didn't see which branch to use as the base in the contributing guide, but it looks like `4-stable` is the default branch now (or always was? probably my mistake). Anyway, I used that as the base branch this time--hope that's alright!

Thanks again for the help!
Thomas


tarebyte:
Thanks @tgmachina ✨

PR #1190, opened 2020-01-30
Fix incorrect API URLs for Apps

ybiquitous:
Hi, I found some incorrect URLs for Apps API. This fixes only URLs in RDoc.


ybiquitous:
Thanks for merging! 😄

PR #1195, opened 2020-02-09
Generate events and checks client

hmharvey:
Part of #1157. Generalizes the `openapi_generator` to work with the events and checks clients and removes hard coded values from the tests. The major changes are replacing `check_run` in method names to just `check`:

### events.rb
`repository_events` => `repo_events`
`organization_events` => `org_events`, also adds user param
`organization_public_events` => `org_public_events`

the same:
`public_events`, `user_events`, `user_public_events`, 

### checks.rb
`create_check_run` => `create_check`
`update_check_run` => `update_check`
`check_runs_for_check_suite` => `check_suites`
`check_run` => `check`
`check_run_annotations` => `check_annotations`
`set_check_suite_preferences` => `set_suite_preferences`

the same:
`check_suite`, `create_check_suite`, `rerequest_check_suite`

Methods removed for now as they'll be generated in other files,
`received_events`, `received_public_events`(user.rb), 
`repository_network_events` (networks.rb)
`repository_issue_events` => `repository_events` (issues.rb)
`issue_events`, `issue_event` (issues.rb)
`check_runs_for_ref`, `check_suites_for_ref` (commits.rb)

### Retroactively:
`move_project_column` and `move_project_card` now return a boolean response


lib/octokit/client/checks.rb
```diff
@@ -1,210 +1,151 @@
...
+      # @param repo [Integer, String, Repository, Hash] A GitHub repository
+      # @param check_suite_id [Integer] The ID of the check suite
+      # @return [Sawyer::Resource] The new suite
+      # @see https://developer.github.com/v3/checks/suites/#rerequest-check-suite
+      def rerequest_check_suite(repo, check_suite_id, options = {})
+        opts = options
+        opts[:accept] = "application/vnd.github.antiope-preview+json" if opts[:accept].nil?
 
-        post "#{Repository.path repo}/check-suites", options
+        post "#{Repository.path repo}/check-suites/#{check_suite_id}/rerequest", opts
```
tarebyte: Should this be a boolean response?

spec/octokit/client/checks_spec.rb
```diff
@@ -3,312 +3,242 @@
...
-          ],
-          summary: "Everything checks out.",
-          title: "Octokit Check",
-        },
-      )
+  context "with commit" do
+    before(:each) do
+      commits = @client.commits(@test_repo)
+      @commit_sha = commits.first.sha
+      @check_name = "octokit-test-check"
```
tarebyte: Thanks for making this not static ✨

spec/openapi_client_generator_spec.rb
```diff
@@ -8,19 +8,21 @@
...
+    #   endpoint = OpenAPIClientGenerator::Endpoint.new(OasParser::Endpoint.new(OasParser::Path.new(nil, nil, {}), "get", json))
+    #   expect(endpoint.singular?).to eq(true)
+    # end
+    #
+    # it "defines singular? as false if response is an array type" do
+    #   json = {"parameters"=> [{"name"=>"owner", "required"=>true}],
+    #          "responses"=> {"200" => {"content"=> {"application/json"=> {"schema"=> {"type"=> "array"}}}}}}
+    #   endpoint = OpenAPIClientGenerator::Endpoint.new(OasParser::Endpoint.new(OasParser::Path.new(nil, nil, {}), "get", json))
+    #   expect(endpoint.singular?).to eq(false)
+    # end
```
tarebyte: Does this need to be uncommented?

lib/octokit/client/checks.rb
hmharvey: I was going to ask you this morning! This endpoint used to always return true, but there are similar endpoints we've done (such as https://developer.github.com/v3/projects/cards/#move-a-project-card) where we just return the response. Which way would we want to generate them?

spec/openapi_client_generator_spec.rb
hmharvey: It doesn't work because we changed how `singular?` is generated and I haven't fixed them yet. I can work on that tho 👍

PR #1197, opened 2020-02-18
Upload release asset

hmharvey:
Part of #1157. The `upload_release_asset` endpoint was recently added as part of the OpenAPI spec, but we're going to include the method definition as an override since generating it was containing a lot of specialized code.


lib/openapi_client_generator.rb
```diff
@@ -338,14 +348,20 @@ def priority
 
   class API
     def self.at(definition, parameterizer: OpenAPIClientGenerator::Endpoint::PositionalParameterizer)
+      overrides_path = "lib/openapi/overrides.rb"
+      source = RuboCop::ProcessedSource.new(File.read(overrides_path), 2.3, overrides_path)
```
hmharvey: We're using Ruby version `2.3` since Rubocop gets bundled at `0.36.0`. I tried forcing it to be higher and it broke some dependencies of other gems so I wasn't sure what we wanted to do.

tarebyte: > I tried forcing it to be higher and it broke some dependencies of other gems so I wasn't sure what we wanted to do.

We can try and take a look at this today.

PR #1199, opened 2020-02-21
Fix API search doc link

NickLaMuro:
Minor change, but based on what I can tell, this should be the proper anchor tag:

https://developer.github.com/v3/search/#search-issues-and-pull-requests

v.s.

https://developer.github.com/v3/search/#search-issues

Which will default to the table of contents (not that a user wouldn't be able to figure it out from there).

**Note**:  Couldn't find it in the docs, but wasn't sure if I should be opening against `4-stable` or `master`.  Choose `master`, but if that is incorrect and I infact can't RTFM, I can re-open this if needed.

### DIFF
--- a/lib/octokit/client/search.rb
+++ b/lib/octokit/client/search.rb
@@ -44,7 +44,7 @@ module Octokit
       # @option options [Integer] :page Page of paginated results
       # @option options [Integer] :per_page Number of items per page
       # @return [Sawyer::Resource] Search results object
-      # @see https://developer.github.com/v3/search/#search-issues
+      # @see https://developer.github.com/v3/search/#search-issues-and-pull-requests
       def search_issues(query, options = {})
         search "search/issues", query, options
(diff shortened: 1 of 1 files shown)


tarebyte:
> Couldn't find it in the docs, but wasn't sure if I should be opening against `4-stable` or `master`. Choose `master`, but if that is incorrect and I infact can't RTFM, I can re-open this if needed.

The `4-stable` branch is the intended branch 😄 we're working on auto generating a new client so we're having folks make contributions against v4

NickLaMuro:
@tarebyte I take that means I should re-open this against `4-stable` then?  😄

tarebyte:
> @tarebyte I take that means I should re-open this against `4-stable` then? 😄

Please

NickLaMuro:
Re-opened as https://github.com/octokit/octokit.rb/pull/1200

PR #1202, opened 2020-02-27
Master

yanz-androidify:
Fixes # #1196


tarebyte:
@yanz-androidify I fail to see how this fixes the linked pr.

PR #1203, opened 2020-02-29
Update applications API use to pass access token as body param instead of path param

oreoshake:
See https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api and https://developer.github.com/changes/2020-02-14-deprecating-oauth-app-endpoint/

This PR updates the calls to the applications API calls. The current implementation uses deprecated endpoint, the new implementation uses the supported, but preview endpoints.

I've got the tests passing but I've yet to actually try this against the live API :smile:


lib/octokit/client/authorizations.rb
```diff
@@ -152,12 +152,13 @@ def scopes(token = @access_token, options = {})
...
       def check_application_authorization(token, options = {})
-        opts = options.dup
+        opts = ensure_api_media_type(:applications_api, options.dup)
+        opts[:access_token] = token
         key    = opts.delete(:client_id)     || client_id
         secret = opts.delete(:client_secret) || client_secret
 
         as_app(key, secret) do |app_client|
-          app_client.get "applications/#{client_id}/tokens/#{token}", opts
+          app_client.post "applications/#{client_id}/tokens", opts
```
augustocravosilva: Hey @tarebyte,
The paths you are using here and returing 404. According to https://developer.github.com/changes/2020-02-14-deprecating-oauth-app-endpoint/#endpoints-affected , it should be singular `token`, and not `tokens`

tarebyte: Damn, thanks for the heads up. I'm working on a PR to resolve this.

oreoshake: > I've got the tests passing but I've yet to actually try this against the live API 😄

🤦‍♂️

tarebyte: https://github.com/octokit/octokit.rb/releases/tag/v4.18.0

augustocravosilva: Thanks for the quick fix. Seems to be working correctly now :)

emma-sax4:
@tarebyte Does this change mean that the newest version of Octokit will address this: https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/?

tarebyte:
> Does this change mean that the newest version of Octokit will address this: https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/?

Yup that should be it!

PR #1218, opened 2020-03-25
Fix incorrect URLs for the Authorizations client

tarebyte:
From https://github.com/octokit/octokit.rb/pull/1203#discussion_r397795550

It turns our we had some bad URLs from https://github.com/octokit/octokit.rb/pull/1203 and so this PR corrects it.

To make things more clear I've moved the methods to a new `OauthApplication` client file and added the old methods as aliases for consistency. I've also rewritten the tests to not use the Authorizations API since it's being deprecated.


tarebyte:
cc @oreoshake

PR #1232, opened 2020-04-03
Regenerate clients

hmharvey:
Part of #1157. This PR refactors the generator a bit and regenerates clients based on the most recent `routes` update. Most method name changes are due to deciding to stick with `repo` instead of manually replacing it for `repository`. This PR also prepares the generator for the next clients, `stats.rb`, `statuses.rb`, and `commits.rb`.


lib/openapi_client_generator.rb
```diff
@@ -145,16 +145,16 @@ def option_overrides
     end
 
     def boolean_response?
-      return true if definition.raw["responses"].key? "204" || definition.raw["responses"].key?("205")
-      return true if definition.raw["responses"].key? "201" && definition.raw["responses"]["201"].keys.count == 1
+      return true if definition.raw["responses"].key? "204" or definition.raw["responses"].key? "205"
+      return true if definition.raw["responses"].key? "201" and definition.raw["responses"]["201"].keys.count == 1
```
hmharvey, line 149: Since `or` and `and` have a lower order of operations I think this is the cleanest way of doing it. Line 149 keeps erring for me with `&&` but we can talk about moving it back if you want 🤔

tarebyte: I think it has to do with the lack of `()`

This worked for me:

```ruby
def boolean_response?
  return true if definition.raw["responses"].key?("204") || definition.raw["responses"].key?("205")
  return true if definition.raw["responses"].key?("201") && definition.raw["responses"]["201"].keys.count == 1

  false
end
```

PR #1233, opened 2020-04-06
Generate stats client

hmharvey:
Part of #1157. Updates the `openapi_cilent_generator` for `stats.rb`. 

The main change here is we're removing the retry logic within Octokit. It's custom code that would be difficult to generate and we felt it best to move it out of the client in favor of using [Octopoller](https://github.com/octokit/octokit.rb#making-repeating-requests).

Method names are all exactly the same 🎉 
- `contributors_stats`, `commit_activity_stats`, `code_frequency_stats`, `participation_stats`, `punch_card_stats`


lib/octokit/client/stats.rb
```diff
@@ -1,106 +1,54 @@
...
-
-      private
-
-      # @private Get stats for a repository
-      #
-      # @param repo [Integer, String, Hash, Repository] A GitHub repository
-      # @param metric [String] The metrics you are looking for
-      # @return [Array<Sawyer::Resource> or nil] Stats in metric-specific format, or nil if not yet calculated.
-      # @see https://developer.github.com/v3/repos/statistics/
-      def get_stats(repo, metric, options = {})
```
tarebyte, line 90: For out future reference, we're going to recommend using https://github.com/octokit/octopoller.rb instead of our built in functionality.

Issue #1216, opened 2020-03-25
Add support for Actions APIs


electrum:
https://developer.github.com/v3/actions/

tarebyte:
@electrum PRs welcome 😄

petar-lazarov:
Hello,
we are going to use part of the GitHub Actions APIs in the near future and it would be easier to have the support.

That's why, I would gladly work on workflow run APIs if that is OK with you :smile:.

Also I noticed there is no endpoint for https://developer.github.com/v3/repos/#create-a-repository-dispatch-event. I will address that too.

tarebyte:
Feel free @petar-lazarov 🍰

ianfixes:
Are there plans to also support the artifacts API within actions?  Or is that also a "PRs welcome" scenario?

tarebyte:
> Or is that also a "PRs welcome" scenario?

Yup 😄

ianfixes:
OK, I'll think about coding that up.  One thing that I've observed is that when downloading artifacts, using `client.get` will attempt to follow a redirect and end up downloading the entire response (which is file contents as a zip) into memory.  That is untenable for large files, in particular docker images (which can potentially be 100s of megabytes if not gigabytes) -- the preferred approach would be to receive an IO stream instead.

What's the most appropriate way to perform the following operation _within_ octokit?
```ruby
require 'net/http'

def artifact_downloadable_url(access_token, owner, repo, run_id, artifact_name)
  client = Octokit::Client.new(access_token: access_token)

  # get URL from artifact
  artifacts = client.get("/repos/#{owner}/#{repo}/actions/runs/#{run_id}/artifacts")[:artifacts]
  artifact = artifacts.find { |a| a[:name] == artifact_name }
  artifact_url = artifact[:archive_download_url]
  
  # extract location header from HTTP 302 response
  uri = URI(artifact_url)
  redirect_response = Net::HTTP.start(uri.host, uri.port, use_ssl: true) do |http|
    request = Net::HTTP::Get.new(uri)
    request["Authorization"] = "token #{access_token}"
    http.request(request)
  end
  redirect_response.each_header.find { |h| h[0] == "location" }[1]
end

# now you can get an IO object as follows:
#   URI(artifact_downloadable_url( ... )).open
```

[...]

PR #1236, opened 2020-04-08
Adds methods for actions secrets

jylitalo:
Still missing all specs for testing these methods, so current status is "works for me".


lib/octokit/client/actions_secrets.rb
```diff
@@ -0,0 +1,53 @@
...
+    #
+    # @see https://developer.github.com/v3/actions/secrets/
+    module ActionsSecrets
+
+      # Get public key for secrets encryption
+      #
+      # @param repo [Integer, String, Hash, Repository] A GitHub repository
+      # @return [Hash] key_id and key
+      def get_public_key(repo)
+        get "#{Repository.path repo}/actions/secrets/public-key"
```
hmharvey, line 14: A couple of really helpful things on all the methods would be adding the developer docs link and an options hash, so it would look something like this:
```suggestion
     # @see https://developer.github.com/v3/actions/secrets/#get-your-public-key
      def get_public_key(repo, options)
        get "#{Repository.path repo}/actions/secrets/public-key", options
```

lib/octokit/client/actions_secrets.rb
```diff
@@ -0,0 +1,53 @@
...
+      end
+
+
+      # List secrets
+      #
+      # @param repo [Integer, String, Hash, Repository] A GitHub repository
+      # @return [Hash] total_count and list of secrets (each item is hash with name, created_at and updated_at)
+      def list_secrets(repo)
+        get "#{Repository.path repo}/actions/secrets"
```
hmharvey, line 23:
```suggestion
        paginate "#{Repository.path repo}/actions/secrets"
```

lib/octokit/client/actions_secrets.rb
```diff
@@ -0,0 +1,53 @@
...
+      def create_or_update_secret(repo, name, options)
+        put "#{Repository.path repo}/actions/secrets/#{name}", options
+      end
+
+      # Delete a secret
+      #
+      # @param repo [Integer, String, Hash, Repository] A GitHub repository
+      # @param name [String] Name of secret
+      def delete_secret(repo, name)
+        delete "#{Repository.path repo}/actions/secrets/#{name}"
```
hmharvey, line 49: To keep it consistent with other endpoints, would you mind changing this to a boolean result?
```suggestion
        boolean_from_response :delete, "#{Repository.path repo}/actions/secrets/#{name}"
```

lib/octokit/client/actions_secrets.rb
jylitalo: Created https://github.com/octokit/octokit.rb/pull/1236/commits/cdd9c749769bc7a997e0d12ffab055ae27174f14 based on this comment.

jylitalo:
On my local machines, this currently fails with:
````
Failures:

  1) Octokit::Client::ActionsSecrets with repository with a secret .create_or_update_secret updating existing secret returns 204
     Got 0 failures and 2 other errors:

     1.1) Failure/Error: raise error

          Octokit::Unauthorized:
            POST https://api.github.com/user/repos: 401 - Bad credentials // See: https://developer.github.com/v3

jylitalo:
According to https://github.com/RubyCrypto/rbnacl/wiki/Installing-libsodium#windows we would need some windows specific stuff into workflow.yaml, so that we get libsolidium.dll into test environment.

pschaumburg:
This solves #1216 right?

jylitalo:
> This solves #1216 right?

I would say that it is only part of the solution, then there are at least artifacts and workflow API calls, which need there own implementations. @see https://developer.github.com/v3/actions/
But the secrets part covers pretty much all that I need to manage via API.

hmharvey:
@jylitalo apologies for the late reply. You're welcome to add to `workflow.yaml` in this PR! And thank you for all the work you've done already 🙇🏻‍♀️

hmharvey:
👋🏻 @jylitalo 

I'm getting access denied when I try to push to this branch, but if you can update it with 4-stable we should be good to merge. We've since removed the dependency on Windows builds  ✨

jylitalo:
Rebased from 4-stable.

jylitalo:
https://github.com/Yleisradio/octokit.rb/actions/runs/119094250 already reports that it passes tests, but for some reason it doesn't yet appear here.

PR #1239, opened 2020-04-10
Generate licenses client

indigok:
Part of #1157. Updates the `openapi_client_generator` for `licenses.rb`.

The method name changes are:
`licenses` => `commonly_used` (the name of the api endpoint has shifted too)
`repository_license_contents` => `repo_license`

And `license` stayed exactly the same


lib/octokit/client/licenses.rb
```diff
@@ -1,43 +1,35 @@
...
-        paginate "licenses", options
+      # @see https://developer.github.com/v3/licenses/#list-commonly-used-licenses
+      def commonly_used(options = {})
+        get 'licenses', options
       end
 
-      # List an individual license
+      # Get an individual license
       #
+      # @param license [String] The license of the license
```
hmharvey, line 20: `The license of the license` isn't a great description but I feel like a better fix than changing our generator would be to rename the parameter in the openapi spec to `license name parameter`. I _think_ this is the preferred style in our openapi guidelines and definitely more consistently with the parameters of other endpoints. But it also seems to take the key of the license which is different from the name of the license as returned by the api, so maybe we should suggest the `license key parameter`? 🤔

hmharvey:
This client generates the same after updating the `openapi_client_generator` with the new endpoint groups from #1237/#1247 🎉

sinstrokes (approved):

sinstrokes (approved):

PR #1242, opened 2020-04-14
Rearrange tests to match generated clients from urls

hmharvey:
This PR updates the test files and cassettes to match the regenerated clients in #1241. Also more broadly a part of #1157.

TODO:
- [ ] double check `comments_spec.rb`
- [ ] double check `reviews_spec.rb`


hmharvey:
Closing this as we're going to prioritize moving tests over to use the octokit [fixtures-server](https://github.com/octokit/fixtures-server).

PR #1252, opened 2020-05-06
Introduce Client methods for Workflows/WorkflowRun

petar-lazarov:
Partially addresses #1216.

:wave:
The PR introduces client methods for ActionWorkflows and ActionWorkflowRuns API endpoints.
Also adds '#dispatch_event' method in Client::Repositories module.


spec/octokit/client/repositories_spec.rb
```diff
@@ -209,6 +209,14 @@
       end
     end # .delete_repository
 
+    describe ".dispath_event", :vcr do
```
hmharvey, line 212:
```suggestion
    describe ".dispatch_event", :vcr do
```

spec/octokit/client/repositories_spec.rb
```diff
@@ -209,6 +209,14 @@
...
+      it "creates a dispatch event" do
+        event_dispatched = @client.dispatch_event(@repo.full_name, 'test dispatch event')
+        expect(event_dispatched).to be_truthy
+        assert_requested :post, github_url("/repos/#{@repo.full_name}/dispatches")
+      end
+    end # .delete_repository
```
hmharvey, line 218:
```suggestion
    end # .dispatch_event
```

petar-lazarov:
Hey @hmharvey, thanks for taking the time to check out the PR 🎉 . I've added a commit, fixing the typos I've done.

lib/octokit/client/actions_workflow_runs.rb
```diff
@@ -0,0 +1,98 @@
...
+      #
+      # @param repo [Integer, String, Repository, Hash] A GitHub repository
+      # @param id [Integer] Id of a workflow run
+      #
+      # @return [Boolean] Returns true if the re-run request was accepted
+      # @see https://developer.github.com/v3/actions/workflow-runs/#re-run-a-workflow
+      def rerun_workflow_run(repo, id, options = {})
+        post "#{Repository.path repo}/actions/runs/#{id}/rerun", options
+
+        last_response.status == 201
```
tarebyte, line 56:
```suggestion
        boolean_from_response :post, "#{Repository.path repo}/actions/runs/#{id}/rerun", options
```

lib/octokit/client/actions_workflow_runs.rb
```diff
@@ -0,0 +1,98 @@
...
+      #
+      # @param repo [Integer, String, Repository, Hash] A GitHub repository
+      # @param id [Integer] Id of a workflow run
+      #
+      # @return [Boolean] Returns true if the cancellation was accepted
+      # @see https://developer.github.com/v3/actions/workflow-runs/#cancel-a-workflow-run
+      def cancel_workflow_run(repo, id, options = {})
+        post "#{Repository.path repo}/actions/runs/#{id}/cancel", options
+
+        last_response.status == 202
```
tarebyte, line 69: Do we want to add the `202` status as an option for `boolean_from_response` so that we could condense this?

```suggestion
        boolean_from_response :post, "#{Repository.path repo}/actions/runs/#{id}/cancel", options
```

petar-lazarov: Hey @tarebyte, thanks for the review :tada:!

I was thinking about making them using `boolean_from_response`. My reasoning to change it this way was because I've seen that boolean_from_response handles `Octokit::NotFound` as false. 

`cancel_workflow_run` raises `Octokit::Conflict` if the run is already completed. `rerun_workflow_run` raises `Octokit::Forbidden` if the run cannot be rerun. Not sure if I have to handle those exception in `boolean_from_response` in like it handles `Octokit::NotFound` and return false in those events.

I think it is going to be useful for the developers to leave those exceptions unhandled because they carry useful information about why the request was not executed correctly.

What do you think, should I handle those exceptions as falsey response or leave them as-is?

tarebyte: > What do you think, should I handle those exceptions as falsey response or leave them unhandled?

I agree the errors here are really useful, I don't think we should rescue them like we do `Octokit::NotFound`.

petar-lazarov: Thanks for the fast response! I've added a commit that addresses the discussion above. :)

PR #1253, opened 2020-05-08
Pass down ssl.verify to Faraday connection

tjwallace:
This ensures adapters properly set SSL related options.

Adapters like Typhoeus [only check `options[:ssl][:verify]`](https://github.com/typhoeus/typhoeus/blob/v1.4.0/lib/typhoeus/adapters/faraday.rb#L137-L141).

### DIFF
--- a/lib/octokit/connection.rb
+++ b/lib/octokit/connection.rb
@@ -178,15 +178,9 @@ module Octokit
       conn_opts = @connection_options
       conn_opts[:builder] = @middleware if @middleware
       conn_opts[:proxy] = @proxy if @proxy
-      if conn_opts[:ssl].nil?
-        conn_opts[:ssl] = { :verify_mode => @ssl_verify_mode } if @ssl_verify_mode
-      else
-        if @connection_options[:ssl][:verify] == false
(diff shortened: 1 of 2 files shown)


lib/octokit/connection.rb
```diff
@@ -175,15 +175,9 @@ def sawyer_options
...
-      else 
-        if @connection_options[:ssl][:verify] == false
-          conn_opts[:ssl] = { :verify_mode => 0}
-        else
-          conn_opts[:ssl] = { :verify_mode => @ssl_verify_mode }
-        end
-      end
+      conn_opts[:ssl] ||= {}
+      conn_opts[:ssl][:verify_mode] ||= 0 if conn_opts.dig(:ssl, :verify) == false
+      conn_opts[:ssl][:verify_mode] ||= @ssl_verify_mode if @ssl_verify_mode
```
tarebyte, line 180: This feels pretty compact, and kind of difficult to read.

Would you mind putting it back to the same formatting as the original?

tjwallace: How about this: fe4a4185?

PR #1270, opened 2020-06-30
Add context with rate limit struct for Octokit::TooManyRequests  and Octokit:: AbuseDetected errors

jatindhankhar:
This adds the `context` attribute to `Octokit:: Error` for much smoother handling of exception like `Octokit::TooManyRequests` and `Octokit:: AbuseDetected`, so that clients consuming them can use the values specified for better retries. 

Can be consumed like this

```ruby
 begin
  octokit_client.get('resource')
rescue Octokit::TooManyRequests => rate_limited_error
  puts "Retry in #{rate_limited_errro.retry_in} seconds"
 # Do fancy sutff
end 
```
Specifically useful when using with background processors like `sidekiq`

```ruby
class PollingWorker
  include Sidekiq::Worker

  sidekiq_retry_in do |attempt, exception|
    case exception
    when Octokit::TooManyRequests
      exception.context.resets_in
    when Octokit::AbuseDetected
      exception.context.resets_in
    end
  end

  def perform
    # Do Polling work
    poller = Octokit::Client.new
    data = poller.fetch('/resource')
  end
end
 
```

The idea is to have meaningful context for all types of errors depending upon the type of error.

### DIFF
--- a/lib/octokit/error.rb
+++ b/lib/octokit/error.rb
@@ -35,8 +35,7 @@ module Octokit
     end
 
      def build_error_context
-       rate_limited_errors = [Octokit::TooManyRequests, Octokit::AbuseDetected]
-       if rate_limited_errors.include?(self.class)
+       if RATE_LIMITED_ERRORS.include?(self.class)
          @context = Octokit::RateLimit.from_response(@response)
        end
      end
@@ -323,4 +322,5 @@ module Octokit
   # Raised when a repository is created with an invalid format
   class InvalidRepository < ArgumentError; end
 
+  RATE_LIMITED_ERRORS = [Octokit::TooManyRequests, Octokit::AbuseDetected]
 end
--- a/lib/octokit/error.rb
+++ b/lib/octokit/error.rb
@@ -1,7 +1,7 @@
 module Octokit
   # Custom error class for rescuing from all GitHub errors
   class Error < StandardError
-
+    attr_reader :context
     # Returns the appropriate Octokit::Error subclass based
     # on status and response message
     #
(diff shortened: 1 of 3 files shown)


lib/octokit/error.rb
```diff
@@ -34,9 +34,17 @@ def self.from_response(response)
       end
     end
 
+     def build_error_context
+       rate_limited_errors = [Octokit::TooManyRequests, Octokit::AbuseDetected]
```
tarebyte, line 38: Can we make this into a constant?

jatindhankhar: Sure, moved the constants to the bottom due to undefined issues.

jatindhankhar:
Thanks for merging this 😄 

I have another proposition that I have been using for my personal projects and work, *custom pagination for Octokit*


## Pagination methods
```ruby
def should_fetch_more?(page_limit, page_iterated)
    page_limit.nil? || page_iterated < page_limit.to_i
  end

  def can_fetch_more?(last_response)
    last_response.rels[:next].present? && @octokit_client.rate_limit.remaining > 0
  end

  def iterate_pagination(opts = {}, result)
    page_limit = opts[:page_limit]
    page_iterated = 1
    last_response = @octokit_client.last_response
    record_limit = opts[:item_limit]
    while should_fetch_more?(page_limit, page_iterated) && can_fetch_more?(last_response)
      last_response = last_response.rels[:next].get
      if result.is_a?(Sawyer::Resource)
        result.items += result.is_a?(Array) ? last_response.data : last_response.data.items
      else
        result += result.is_a?(Array) ? last_response.data : last_response.data.items
      end

      page_iterated += 1
      return limit_result_count(result, record_limit.to_i) if record_limit && result.count >= record_limit.to_i
    end
    record_limit ? limit_result_count(result, record_limit.to_i) : result
  end
```

so any utility function can be passed as proc to `execute`

```ruby
```

```ruby
def find_users(search_text = '', opts = {})
    execute(opts) { @octokit_client.search_users(search_text).items }
end

```

tarebyte:
@jatindhankhar mind opening another issue? It's hard to track requests at the end of a merged PR.

PR #1287, opened 2020-09-15
Add search topics

xzdshr:
Add missing endpoint for the topics search
https://docs.github.com/en/rest/reference/search#search-topics


lib/octokit/client/search.rb
```diff
@@ -64,6 +64,18 @@ def search_repositories(query, options = {})
...
+      # Search topics
+      #
+      # @param query [String] Search term and qualifiers
+      # @param options [Hash] Pagination options
+      # @option options [Integer] :page Page of paginated results
+      # @option options [Integer] :per_page Number of items per page
+      # @return [Sawyer::Resource] Search results object
+      # @see https://developer.github.com/v3/search/#search-topics
+      def search_topics(query, options = {})
+        search "search/topics", query, options
```
tarebyte, line 76: It looks like we're missing the `application/vnd.github.mercy-preview+json` preview header for this request.

tarebyte (changes requested):
Thanks for opening this up! Just one bit of feedback

lib/octokit/client/search.rb
xzdshr: totally missed it, thanks!

tarebyte (approved):

PR #1297, opened 2020-10-28
[WIP] Escape label names with URL characters

### LINKED ISSUE #1296: If a label contains a `?`, update_label silently fails
In our repository, we have a label `jansa/yes?`...If I try to call

```ruby
client.update_label(repo, label, :color => color)
```

then it silently fails, but if I call it with

```ruby
require "cgi"
client.update_label(repo, CGI.escape(label), :color => color)
```

then it works.

I can't tell if this is an issue with octokit, sawyer, or the GitHub API itself.  From what I can tell, the label and color are passed as data and so should be the body of a PATCH request, which makes me think this is GitHub itself?

Fryguy:
Fixes #1296

@tarebyte This is WIP because I am having trouble recording the VCR cassettes, and I'm not sure what I'm doing wrong.  I could probably use some pointers there.


Fryguy:
The issue I'm hitting with VCR specs seems to be that it can't create the test repository.

First thing I ran into is that faraday 1.0 doesn't seem to work with this repo since they changed the namespacing of their error class, so I had to lock that down to "< 1.0" in the gemspec. (I'll PR this change separately)

Now I'm hitting the following which is happening when it's trying to create the test repository, but I can't figure out why or how to get past it.

```
  1) Octokit::Client::Labels with issue .remove_label removes a label from the specified issue
     Failure/Error: @issue = @client.create_issue(@test_repo, "Issue for label test", "The body")
     NoMethodError:
       undefined method `<<' for #<Hash:0x00007f91c7486c90>
       Did you mean?  <
     # ./lib/octokit/middleware/follow_redirects.rb:73:in `perform_with_redirection'
     # ./lib/octokit/middleware/follow_redirects.rb:61:in `call'
     # ./lib/octokit/connection.rb:156:in `request'
     # ./lib/octokit/connection.rb:19:in `get'
     # ./lib/octokit/client/repositories.rb:29:in `repository'
     # ./lib/octokit/client/repositories.rb:15:in `repository?'
     # ./spec/helper.rb:89:in `block (2 levels) in <top (required)>'
     # ./lib/octokit/middleware/follow_redirects.rb:73:in `perform_with_redirection'
     # ./lib/octokit/middleware/follow_redirects.rb:61:in `call'
     # ./lib/octokit/connection.rb:156:in `request'
     # ./lib/octokit/connection.rb:28:in `post'
     # ./lib/octokit/client/issues.rb:101:in `create_issue'
     # ./spec/octokit/client/labels_spec.rb:66:in `block (3 levels) in <top (required)>'
```

Fryguy:
> First thing I ran into is that faraday 1.0 doesn't seem to work with this repo since they changed the namespacing of their error class, so I had to lock that down to "< 1.0" in the gemspec. (I'll PR this change separately)

I didn't realize this repo didn't use the usual `master` branch but instead used `4-stable`, which already has this faraday thing fixed.

It seems this may also be the solution for the VCR issues I'm having, so now working through rerecording those.

Fryguy:
@tarebyte This is ready to go!

sinstrokes (approved):

Fryguy:
@tarebyte Is this good to merge?

Fryguy:
@tarebyte Are you good with this change?

Fryguy:
@tarebyte @hmharvey Bump?

Fryguy:
Rebased against 4-stable to verify on the new 3.0 test suite/

@tarebyte Is there any else you need me to do here to get this in?  Should it go against a different branch or multiple branches?

PR #1302, opened 2020-11-07
Add documentation on how to specify the ref option for RubyDoc

aomathwift:
When you pass the ref option, you may not notice the syntax [here](https://github.com/octokit/octokit.rb/blob/8f95de13cb09cdbbf71e79d533ca52e009021554/README.md#additional-query-parameters). 
Therefore, it is convenient if this syntax is also described in RubyDoc.


rtrtrt777 (approved):
set update

rtrtrt777:
set

newwwef90 (approved):

tarebyte (approved):

PR #1304, opened 2020-11-12
Added the ability to delete a deployment

jer-k:
Hello, I recently started working on the ability for my company's application to use the Deployments API and I noticed that there was no way to delete the Deployments through Octokit. We make new environments on every Pull Request so without the ability to delete the Deployments, our Environments section is going to become overrun. I was hoping to add this functionality to Octokit rather than having to hit the Github API directly.


tarebyte (approved):
Looks great, thanks so much!

PR #1321, opened 2021-01-28
Paginate when retrieving check runs and check suites

a2ikm:
As far as I've investigated, the following check APIs return the link headers:

- [List check runs in a check suite](https://docs.github.com/en/rest/reference/checks#list-check-runs-in-a-check-suite)
- [List check runs for a Git reference](https://docs.github.com/en/rest/reference/checks#list-check-runs-for-a-git-reference)
- [List check suites for a Git reference](https://docs.github.com/en/rest/reference/checks#list-check-suites-for-a-git-reference)

This PR allows `auto_paginate` in them.

### DIFF
--- a/lib/octokit/client/checks.rb
+++ b/lib/octokit/client/checks.rb
@@ -65,7 +65,9 @@ module Octokit
       def check_runs_for_ref(repo, ref, options = {})
         ensure_api_media_type(:checks, options)
 
-        get "#{Repository.path repo}/commits/#{ref}/check-runs", options
+        paginate "#{Repository.path repo}/commits/#{ref}/check-runs", options do |data, last_response|
+          data.check_runs.concat last_response.data.check_runs
+        end
       end
       alias :list_check_runs_for_ref :check_runs_for_ref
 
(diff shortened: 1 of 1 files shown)


tarebyte (changes requested):
Thanks for the contribution!

We let the client decide whether they want to auto paginate or not in their settings https://github.com/octokit/octokit.rb#pagination. This helps with rate limiting an other factors.

Would you mind just updating the calls to use `paginate` without the block?

a2ikm:
Thanks for the comment!
Yes, that's exactly why I added the block 😄 

> Mind just adding tests 😅

Okay, I'll add tests 👍 
It may take a time because we should update VCR cassetts to include two or more documents in the lists.

lastthyme (approved):

matiasalbarello (approved):

lhmzhou (approved):

kou:
@a2ikm Do you still want to add tests for this changes?

@matiasalbarello @lhmzhou Do you want to help it?

Smiley278:
Yes

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

a2ikm:
@kou 
Sorry for my too late reaction.
I have no time to add tests.

On the other hand, @matiasalbarello's PR looks good.
So please continue with it 🙏

nickfloyd:
Closing this one in favor of: @matiasalbarello's [PR](https://github.com/octokit/octokit.rb/pull/1415).  Thank you @a2ikm for your effort here and for getting this work in motion.  ❤️

PR #1322, opened 2021-02-02
Add matching refs

AHaymond:
Adding the ability to retrieve matching refs based on suffix per Github API V3 [docs](https://docs.github.com/en/rest/reference/git#list-matching-references)


lib/octokit/client/refs.rb
```diff
@@ -23,6 +23,19 @@ def refs(repo, namespace = nil, options = {})
...
+      #
+      # @param repo [Integer, String, Repository, Hash] A GitHub repository
+      # @param ref [String] The ref, e.g. <tt>tags/v0.0.3</tt> or <tt>heads/rails-3</tt>
+      # @return [Array<Sawyer::Resource>] The reference matching the given repo and the ref id
+      # @see https://developer.github.com/v3/git/refs/#list-matching-references
+      # @example Fetch refs matching  tags/v2 for sferik/rails_admin
+      #   Octokit.ref("sferik/rails_admin","tags/v2")
+      def matching_refs(repo, ref, options = {})
+        path = "#{Repository.path repo}/git/matching-refs/#{ref}"
+        paginate path, options
```
tarebyte, line 36: Since we aren't doing anything fancy to the path, we can probably just leave it all as one line:

```suggestion
        paginate "#{Repository.path repo}/git/matching-refs/#{ref}", options
```

tarebyte (approved):

lib/octokit/client/refs.rb
AHaymond: Absolutely, thanks for the suggestion.

PR #1329, opened 2021-02-25
Rename branch support

### LINKED ISSUE #1328: Branch renaming support
I see support for branch renames in the [JS client](https://octokit.github.io/rest.js/v18#repos-rename-branch), but not here in the Ruby client. Are there any plans to add support for this? If there hasn't been any work done yet, I'd be happy to help!

Note that I'm not referring to _setting_ a default branch. I'm referring to the new `rename` feature that auto-switches PR targets, inherits protection rules, etc.

gmcgibbon:
Closes #1328 

https://docs.github.com/en/rest/reference/repos/#rename-a-branch

This adds support for the renaming branches to the Ruby client. ~We still need a VCR cassette that works. Would using my own repo with temp access token break tests?~

cc @tarebyte


gmcgibbon:
Tests are 🟢! I see the obfuscation in the cassette, but deleted the API token I used, just to be safe. Please review this when you can, @tarebyte ❤️

tarebyte (approved):

PR #1332, opened 2021-04-07
Added the ability to get GitHub Actions billing for an organization.

M-Yamashita01:
Hello, we recently started developing tools to report usage and billing every month for GitHub Actions with Octokit.
During development, we noticed that there was no way to get the billing through Octokit.
Get GitHub Actions Billing for an organization: https://docs.github.com/en/rest/reference/billing#get-github-actions-billing-for-an-organization

Therefore, we hope to add this functionality to Octokit rather than having to hit the GitHub API directly.

Let me know if anything is unclear, or could be improved. Thank you.


hmharvey (approved):
Thank you for adding this ✨

PR #1335, opened 2021-05-04
Update README to remove Octokit client instance creation with username and password.

RaminMammadzada:
As of November 2020, the GitHub API does not allow you to authenticate to the API via username and password. See https://github.blog/changelog/2020-11-13-token-authentication-required-for-api-operations/ for more details.

Therefore login with user and password must be removed from the README file of the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,10 +86,8 @@ Access the library in Ruby:
 
 ```ruby
 # Provide authentication credentials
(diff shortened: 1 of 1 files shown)


hmharvey:
👋🏻 thanks for the contribution! However, the branch `4-stable` is the default branch and the one we're adding improvements on, which I believe already has this update.

Feel free to reopen on `4-stable` if it doesn't though! ✨

PR #1338, opened 2021-05-06
Update README to remove Octokit client instance creation with username and password.

RaminMammadzada:
As of November 2020, the GitHub API does not allow you to authenticate to the API via username and password. See https://github.blog/changelog/2020-11-13-token-authentication-required-for-api-operations/ for more details.

Therefore login with user and password must be removed from the README file of the project.

PR #1341, opened 2021-05-18
Revert "Update README to remove Octokit client instance creation with username and password."

hmharvey:
Reverts octokit/octokit.rb#1338

Although the username/password authentication has been deprecated, you can use the syntax as long as you [use your PAT](https://docs.github.com/en/rest/overview/other-authentication-methods#via-oauth-and-personal-access-tokens) instead of your password. Thus, these comments were valid so we can add them back ✨


tarebyte (approved):

mspaansen (commented):
Ok

PR #1351, opened 2021-07-22
Make clients autoload

gmcgibbon:
Make clients autoload so that requiring the gem doesn't slow down boot time of apps that use this gem. In our app, it took about 400ms to load `Octokit::Client` on boot:

Screen Shot 2021-07-22 at 3 41 36 PM

You can test this on a simpler scale by wrapping the test suite require in a benchmark block like so:
```ruby
# spec/helper.rb

require "benchmark"
ms = Benchmark.realtime do
  require 'octokit'
end * 1000
puts "took #{ms}ms to laod"
```

Pre patch this took around ~200ms and post patch about 110ms on my machine. Not quite 2x faster, but still worth autoloading I think. The autoload patched client also removes octokit from our app profiles, which means the require time is reduced significantly in both cases.


tas50:
@gmcgibbon What are you using to visualize the load time there?

gmcgibbon:
@tas50 I'm using [stackprof](https://github.com/tmm1/stackprof) to measure and [speedscope](https://github.com/jlfwong/speedscope) to visualize!

lib/octokit/client.rb
```diff
@@ -2,7 +2,6 @@
 require 'octokit/warnable'
 require 'octokit/arguments'
 require 'octokit/repo_arguments'
-require 'octokit/configurable'
 require 'octokit/authentication'
 require 'octokit/gist'
 require 'octokit/rate_limit'
```
jhawthorn, line 7: Should we move these to autoloads in the `Octokit` module as well in case they are referenced before `Client`?

gmcgibbon: Right... it depends on the API of the gem. It looks like they are all public API and you expect these modules to be pluggable by other client types. 

I'll take a look at the modules that aren't under `Client` and see if it makes sense to autoload them. If they are trivial enough, we could consider just moving the require.

gmcgibbon: I decided to autoload all constants nested in `Octokit` besides `Configurable` (because that's used immediately in the main module). It should be roughly the same performance diff that way.

jhawthorn (approved):
Looks great to me!

gmcgibbon:
@jhawthorn are you able to ping someone with merge access to look at this? It looks like this is still an issue.

nickfloyd:
@gmcgibbon we're needing to roll back this change in our next release.  It turns out this had some side effects that we are going to need to think through and work around.  see https://github.com/octokit/octokit.rb/issues/1427

Apologies for the back and forth on this - the change you proposed here is a good one, we just need to introduce it in such a  way that doesn't cause harm to other users.

nickfloyd:
Reverted in release [v4.24.0](https://github.com/octokit/octokit.rb/releases/tag/v4.24.0)

PR #1354, opened 2021-08-11
CI: Use RUBYOPT: "--enable-frozen-string-literal" to find any incompatibilities

olleolleolle:
This PR makes a try-out with the "frozen string literals" setting.

If it works, then great, if not, we learned something. **Update**: We learned something.

Now, the **only failures that are left** come from the `multipart-post` gem, and its next release will fix those, too.

This PR **fixes** the parts in **octokit.rb** which used strings in incompatible ways.

---

## Changes

- build_error_message could use a mutable string 4dcea1c, 9f9c7a0
- CI: use an updated version (still 2.x) of RubyGems for Ruby 2.5

## Learnings

- ✅ Ruby 2.5's old-but-shippped rubygems version had an issue - https://github.com/rubygems/rubygems/issues/4837 - so we could just update away from that problem for 2.5.
- ✅ the `multipart-post` gem had a frozen string literal issue, now fixed https://github.com/socketry/multipart-post/pull/70


olleolleolle:
I learned that one file in RubyGems is modifying a string in-place:
https://github.com/rubygems/rubygems/blob/master/lib/rubygems/bundler_version_finder.rb#L87

timrogers:
Sadly, it looks like the [fix](https://github.com/socketry/multipart-post/pull/70) on `multipart-post` still hasn't been released into a version, so this is stuck until then. I'll try creating an issue over there.

PR #1356, opened 2021-08-15
Fixes grammar and style

nikoandpiko:
Grammar issues are of most importance here. If anything needs updating, I would suggest those.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ When passing additional parameters to GET based request use the following syntax
  client.contents('octokit/octokit.rb', path: 'path/to/file.rb', query: {ref: 'some-other-branch'})
 ```
 
-[API methods]: http://octokit.github.io/octokit.rb/method_list.html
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -109,7 +109,7 @@ When passing additional parameters to GET based request use the following syntax
  client.contents('octokit/octokit.rb', path: 'path/to/file.rb', query: {ref: 'some-other-branch'})
 ```
 
-[API methods]: http://octokit.github.io/octokit.rb/method_list.html
+[api methods]: http://octokit.github.io/octokit.rb/method_list.html
```
nickfloyd, line 112: nit: API should be in caps given it is an acronym - definitely not a show stopper for the great changes in this PR though

```suggestion
[API methods]: http://octokit.github.io/octokit.rb/method_list.html
```

nickfloyd (approved):
Thank you for making these changes ❤️ .  Just one small nit but definitely not a blocker.

PR #1358, opened 2021-08-31
Fix Faraday required minimum version

ybiquitous:
This change increases the Faraday required minimum version from `0.9` to `0.16.2`.

When running the tests with Faraday 0.15, we get the following error:

```
TypeError: superclass mismatch for class RequestOptions
```

This change also adds a `FARADAY_VERSION` environment variable and a Faraday matrix to the CI configuration.
By this addition, you can run the tests with any Faraday version like this:

```shell
FARADAY_VERSION='~> 0.16.0' bundle install && bundle exec rspec
```

See also [all Faraday's versions](https://rubygems.org/gems/faraday/versions)

### DIFF
--- a/.github/workflows/octokit.yml
+++ b/.github/workflows/octokit.yml
@@ -13,7 +13,7 @@ on:
 
 jobs:
   ci:
-    name: Ruby ${{ matrix.ruby }}
+    name: 'Ruby ${{ matrix.ruby }} with Faraday ${{ matrix.faraday }}'
     runs-on: ${{ matrix.os }}-latest
 
     strategy:
@@ -21,6 +21,10 @@ jobs:
       matrix:
         os: [ ubuntu ]
         ruby: [ 2.5, 2.6, 2.7, '3.0', head ]
+        faraday: [ '~> 0.16.0', '~> 0.17.0', '>= 1.0' ]
+
+    env:
+      FARADAY_VERSION: ${{ matrix.faraday }}
 
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


ybiquitous:
[Faraday 0.16.0](https://github.com/lostisland/faraday/releases/tag/v0.16.0) seems to include 1.0-RC changes. Perhaps, it may be better to change to `>= 1.0`, instead of `>= 0.16.2`. 🤔

.github/workflows/octokit.yml
```diff
@@ -13,24 +13,28 @@ on:
...
-    name: Ruby ${{ matrix.ruby }}
+    name: 'Ruby ${{ matrix.ruby }} with Faraday ${{ matrix.faraday }}'
     runs-on: ${{ matrix.os }}-latest
 
     strategy:
       fail-fast: false
       matrix:
         os: [ ubuntu ]
         ruby: [ 2.5, 2.6, 2.7, head ]
+        faraday: [ '~> 0.16.0', '~> 0.17.0', '>= 1.0' ]
```
ybiquitous, line 24: [note] `~> 1.0` may be better?

```suggestion
        faraday: [ '~> 0.16.0', '~> 0.17.0', '~> 1.0' ]
```

octokit.gemspec
```diff
@@ -6,7 +6,7 @@ require 'octokit/version'
 Gem::Specification.new do |spec|
   spec.add_development_dependency 'bundler', '>= 1', '< 3'
   spec.add_dependency 'sawyer', '>= 0.5.3', '~> 0.8.0'
```
tisba, line 8: sawyer 0.9.0 was released, which is blocking from updating to faraday 2.

The only changes betwen sawyer 0.8.0 and 0.9.0 are https://github.com/lostisland/sawyer/pull/67 and https://github.com/lostisland/sawyer/pull/72.

```suggestion
  spec.add_dependency 'sawyer', '>= 0.5.3', '~> 0.9.0'
```

ybiquitous: @tisba Thanks for sharing the new release. 👍🏼

tisba:
Anything I/we can do to move this forward? Would love to finally upgrade to faraday 2 :)

ybiquitous:
@tisba It may be better to create a new pull request because there is a big difference between HEAD and this branch. So, if you're interested, I'll close this PR. 😃

ybiquitous:
@nickfloyd No problems. Thanks for getting in touch!

PR #1359, opened 2021-08-31
Fix deprecation warning since Faraday 1.7.1

### LINKED ISSUE #1357: Deprecation warning since Faraday 1.7.1
Since [Faraday 1.7.1](https://github.com/lostisland/faraday/releases/tag/v1.7.1), the following methods of the `Faraday::Connection` class emits a deprecation warning:

- `#basic_auth`
- `#token_auth`
- `#authorization`

See also PR lostisland/faraday#1306

When running the specs on my local machine with the current HEAD revision (605a08a2), we can see the many warnings:

```console
$ bundle exec rspec
...
WARNING: `Faraday::Connection#authorization` is deprecated; it will be removed in version 2.0.
While initializing your connection, use `#request(:authorization, ...)` instead.
See https://lostisland.github.io/faraday/middleware/authentication for more usage info.
...
```

Used gem versions are as below:

```console
$ bundle list | grep faraday
  * faraday (1.7.1)
  * faraday-em_http (1.0.0)
  * faraday-em_synchrony (1.0.0)
  * faraday-excon (1.1.0)
  * faraday-httpclient (1.0.1)
  * faraday-net_http (1.0.1)
  * faraday-net_http_persistent (1.2.0)
  * faraday-patron (1.0.0)
  * faraday-rack (1.0.0)
```

I think Octokit needs to fix the warnings to prevent users' application logs from being flooded with them.

ybiquitous:
Close #1357

### DIFF
--- a/lib/octokit/client.rb
+++ b/lib/octokit/client.rb
@@ -240,7 +240,7 @@ module Octokit
         if basic_authenticated?
           http.request :basic_auth, @login, @password
         elsif token_authenticated?
-          http.request :token_auth, @access_token
+          http.request :authorization, 'token', @access_token
         elsif bearer_authenticated?
           http.request :authorization, 'Bearer', @bearer_token
         end
--- a/lib/octokit/client.rb
+++ b/lib/octokit/client.rb
@@ -238,11 +238,11 @@ module Octokit
       conn_opts[:ssl] = { :verify_mode => @ssl_verify_mode } if @ssl_verify_mode
       conn = Faraday.new(conn_opts) do |http|
         if basic_authenticated?
-          http.basic_auth(@login, @password)
+          http.set_authorization_header :basic_auth, @login, @password
         elsif token_authenticated?
-          http.authorization 'token', @access_token
+          http.set_authorization_header :authorization, 'token', @access_token
         elsif bearer_authenticated?
-          http.authorization 'Bearer', @bearer_token
+          http.set_authorization_header :authorization, 'Bearer', @bearer_token
         end
         http.headers['accept'] = options[:accept] if options.key?(:accept)
       end
--- a/lib/octokit/client.rb
+++ b/lib/octokit/client.rb
@@ -238,11 +238,11 @@ module Octokit
       conn_opts[:ssl] = { :verify_mode => @ssl_verify_mode } if @ssl_verify_mode
       conn = Faraday.new(conn_opts) do |http|
         if basic_authenticated?
-          http.set_authorization_header :basic_auth, @login, @password
+          http.request :basic_auth, @login, @password
         elsif token_authenticated?
-          http.set_authorization_header :authorization, 'token', @access_token
+          http.request :token_auth, @access_token
         elsif bearer_authenticated?
-          http.set_authorization_header :authorization, 'Bearer', @bearer_token
+          http.request :authorization, 'Bearer', @bearer_token
         end
         http.headers['accept'] = options[:accept] if options.key?(:accept)
       end
--- a/lib/octokit/client/pub_sub_hubbub.rb
+++ b/lib/octokit/client/pub_sub_hubbub.rb
@@ -91,9 +91,9 @@ module Octokit
         conn = Faraday.new(:url => @api_endpoint) do |http|
           http.headers[:user_agent] = user_agent
           if basic_authenticated?
-            http.set_authorization_header :basic_auth, @login, @password
+            http.request :basic_auth, @login, @password
           elsif token_authenticated?
-            http.set_authorization_header :authorization, 'token', @access_token
+            http.request :token_auth, @access_token
           end
           http.request  :url_encoded
           http.use Octokit::Response::RaiseError
--- a/lib/octokit/client/pub_sub_hubbub.rb
+++ b/lib/octokit/client/pub_sub_hubbub.rb
@@ -93,7 +93,7 @@ module Octokit
           if basic_authenticated?
             http.request :basic_auth, @login, @password
           elsif token_authenticated?
-            http.request :token_auth, @access_token
+            http.request :authorization, 'token', @access_token
           end
           http.request  :url_encoded
           http.use Octokit::Response::RaiseError
--- a/lib/octokit/client/pub_sub_hubbub.rb
+++ b/lib/octokit/client/pub_sub_hubbub.rb
@@ -91,9 +91,9 @@ module Octokit
         conn = Faraday.new(:url => @api_endpoint) do |http|
           http.headers[:user_agent] = user_agent
           if basic_authenticated?
-            http.basic_auth(@login, @password)
+            http.set_authorization_header :basic_auth, @login, @password
           elsif token_authenticated?
-            http.authorization 'token', @access_token
+            http.set_authorization_header :authorization, 'token', @access_token
           end
           http.request  :url_encoded
           http.use Octokit::Response::RaiseError
--- a/lib/octokit/connection.rb
+++ b/lib/octokit/connection.rb
@@ -176,7 +176,7 @@ module Octokit
         :links_parser => Sawyer::LinkParsers::Simple.new
       }
       conn_opts = @connection_options
-      conn_opts[:builder] = @middleware if @middleware
+      conn_opts[:builder] = @middleware.dup if @middleware
       conn_opts[:proxy] = @proxy if @proxy
       if conn_opts[:ssl].nil?
         conn_opts[:ssl] = { :verify_mode => @ssl_verify_mode } if @ssl_verify_mode
--- a/lib/octokit/connection.rb
+++ b/lib/octokit/connection.rb
@@ -109,7 +109,7 @@ module Octokit
         if basic_authenticated?
           http.request :basic_auth, @login, @password
         elsif token_authenticated?
-          http.request :token_auth, @access_token
+          http.request :authorization, 'token', @access_token
         elsif bearer_authenticated?
           http.request :authorization, 'Bearer', @bearer_token
         elsif application_authenticated?
--- a/lib/octokit/connection.rb
+++ b/lib/octokit/connection.rb
@@ -107,13 +107,13 @@ module Octokit
         http.headers[:content_type] = "application/json"
         http.headers[:user_agent] = user_agent
         if basic_authenticated?
-          http.basic_auth(@login, @password)
+          http.set_authorization_header :basic_auth, @login, @password
         elsif token_authenticated?
-          http.authorization 'token', @access_token
+          http.set_authorization_header :authorization, 'token', @access_token
         elsif bearer_authenticated?
-          http.authorization 'Bearer', @bearer_token
+          http.set_authorization_header :authorization, 'Bearer', @bearer_token
         elsif application_authenticated?
-          http.basic_auth(@client_id, @client_secret)
+          http.set_authorization_header :basic_auth, @client_id, @client_secret
         end
       end
     end
--- a/lib/octokit/connection.rb
+++ b/lib/octokit/connection.rb
@@ -107,13 +107,13 @@ module Octokit
         http.headers[:content_type] = "application/json"
         http.headers[:user_agent] = user_agent
         if basic_authenticated?
-          http.set_authorization_header :basic_auth, @login, @password
+          http.request :basic_auth, @login, @password
         elsif token_authenticated?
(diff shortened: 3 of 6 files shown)


sambostock (changes requested):
It looks like `set_authorization_header` was also removed upstream in https://github.com/lostisland/faraday/pull/1308

Based on the deprecation message and diff in that PR, and adjusting for consistency, I think we want:

```diff
-http.basic_auth                       "username", "password"
+http.request :authorization, "Basic", "username", "password"
```

```diff
-http.authorization           "token", "access-token"
+http.request :authorization, "Token", "access-token"
```

```diff
-http.authorization           "Bearer", "bearer-token"
+http.request :authorization, "Bearer", "bearer-token"
```

sambostock:
Actually, it looks like

```ruby
http.request :authorization, "Basic", "username", "password"
```

isn't **backwards** compatible with Faraday versions prior to that PR. It blows up because of the extra password argument.

One forward **and** backward compatible way of handling it would be to encode the header ourselves

```ruby
require "base64"
http.request :authorization, "Basic", Base64.strict_encode64("#{username}:#{password}")
```

ybiquitous:
@sambostock Thanks for the feedback. Yes, this PR is not backward compatible with Faraday v2.

The PR aims to suppress the annoying warnings, and I think we may need additional code changes for Faraday v2.

etiennebarrie (commented):
`Faraday::Connection#set_authorization_header` has been removed from main, so it's not a long-term fix.

olleolleolle:
Would it be fruitful to try with  Faraday 1.8.0, which came in the meantime?

ybiquitous:
@olleolleolle Thanks for sharing the new version. This PR seems still valid with Faraday 1.8.0.

olleolleolle:
@ybiquitous Yeah, this PR 👍 excellent stuff. 1.0 + 2.0 compat may take another PR, and some trickery.

iMacTia:
Hi all, just a quick nudge to let you know that we've added a ["how to do this in Faraday 1.x" section](https://lostisland.github.io/faraday/middleware/authentication#faraday-1x-usage) in the documentation.
This should help solving the warning messages, which I understand is the objective of this PR.

Regarding cross-supporting both Faraday 1.x and 2.x, I'm afraid at the moment there's no suggested way of doing it.
The solution proposed [here](https://github.com/octokit/octokit.rb/pull/1359#issuecomment-916389545) should work, or alternatively you could always test the Faraday version and use the correct middleware based on that.

I'm open to feedback/suggestions if you have any, but I agree this should be tackled on a separate PR

ybiquitous:
@iMacTia Thank you. I've followed your advice and push the change c34a880b, but the tests have failed:

```
Faraday::RackBuilder::StackLocked:
  can't modify middleware stack after making a request
```

https://github.com/octokit/octokit.rb/pull/1359/checks?check_run_id=3673807356#step:7:3522

Any advice?

iMacTia:
@ybiquitous I was having a look myself and I'm finding it quite hard to understand what's going on.
The error is due to the fact that once the first request is fired on a connection, the underlying builder is locked, so that calling `http.request` is not allowed anymore.

However, looking at the code pointed by the stacktrace I can see that a new connection is being initialised, with no request being fired until the end of the process. So I don't really understand how that can be possible.

So far, the only thing I can think about is that [this line](https://github.com/octokit/octokit.rb/pull/1359/files#diff-93659c33383fb03685f062cf7b8a13600c7df68c4ef8106855fb362c40b4da54L179) is somehow injecting an already locked builder, but I'm having some really hard time following the code (and in particular usages of `@middleware` given the magic/hidden way these variables are injected 🤔

ybiquitous:
@iMacTia Thanks for the investigation. Maybe, the `MIDDLEWARE` constant seems to be a failure cause.

https://github.com/octokit/octokit.rb/blob/d729c0dfacf4aea854ac4899471bbbf59aa33035/lib/octokit/default.rb#L26-L33

The following change can reduce [556 failures](https://github.com/octokit/octokit.rb/pull/1359/checks?check_run_id=3673807494#step:7:12656) to **46 failures** on my localhost:

```diff
--- a/lib/octokit/default.rb
+++ b/lib/octokit/default.rb
@@ -115,7 +115,7 @@ module Octokit
       # from {MIDDLEWARE}
       # @return [Faraday::RackBuilder or Faraday::Builder]
       def middleware
-        MIDDLEWARE
+        MIDDLEWARE.dup
       end
 
       # Default GitHub password for Basic Auth from ENV
```

(Note: this change is **not** an essential solution)

`Faraday::RackBuilder::StackLocked` still occurs...

iMacTia:
Nice catch @ybiquitous 👏 ! It looks like a step in the right direction though. Somehow that same connection is being reused, causing new agents to fail after the first request gets processed 👍 
Can you push the change that for works locally so we can check the remaining failing tests?

ybiquitous:
Sure. I've pushed 1680badb.

iMacTia:
Thanks @ybiquitous, that was indeed useful, I think I found the underlying issue:
https://github.com/octokit/octokit.rb/blob/d1628ebd3e71aeb1534b3115c8bf37e8fb42068f/lib/octokit/client.rb#L135-L138

The Client initialiser sets instance variables based on the main Octokit module variables. This is quite convoluted, but `middleware` is one of those. Your last commit only fixes the case where `@middleware` is accessed through the getter method, but in this case each instance variables is accessed directly

ybiquitous:
@iMacTia Thank you so much! I've added `@middleware.dup` (9fadd8bb) also to `Octokit::Connection` like `Octokit::Client`:

https://github.com/octokit/octokit.rb/blob/d729c0dfacf4aea854ac4899471bbbf59aa33035/lib/octokit/client.rb#L236

iMacTia:
Only 4 failures left 🎉!
We need to get more information from VCR, you should be able to do so using the [`debug-logger`](https://relishapp.com/vcr/vcr/v/5-1-0/docs/configuration/debug-logging)

ybiquitous:
@iMacTia Thank you. When I turned on the Faraday's logging as below,

```diff
--- a/lib/octokit/connection.rb
+++ b/lib/octokit/connection.rb
@@ -106,6 +106,7 @@ module Octokit
         http.headers[:accept] = default_media_type
         http.headers[:content_type] = "application/json"
         http.headers[:user_agent] = user_agent
+        http.response :logger
         if basic_authenticated?
           http.request :basic_auth, @login, @password
         elsif token_authenticated?
```

I got the following result. When using `http.request :token_auth`, the `Authorization` header is not set.
Is there something wrong with the usage of `http.request`?

---

Using `http.authorization 'token', @access_token` (before):

```
I, [2021-09-23T00:25:08.854295 #4998]  INFO -- request: GET https://api.github.com/
I, [2021-09-23T00:25:08.854330 #4998]  INFO -- request: Accept: "application/vnd.github.v3+json"
User-Agent: "Octokit Ruby Gem 4.21.0"
Content-Type: "application/json"
Authorization: "token xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
I, [2021-09-23T00:25:08.859670 #4998]  INFO -- response: Status 200
```

Using `http.request :token_auth, @access_token` (after):

```
I, [2021-09-23T00:25:27.288127 #5129]  INFO -- request: GET https://api.github.com/
I, [2021-09-23T00:25:27.288161 #5129]  INFO -- request: Accept: "application/vnd.github.v3+json"
User-Agent: "Octokit Ruby Gem 4.21.0"
Content-Type: "application/json"
I, [2021-09-23T00:25:27.295578 #5129]  INFO -- response: Status 401
```

iMacTia:
@ybiquitous it's an issue with middleware order. When you use the deprecated `http.authorization`, you're effectively setting a default header on the connection, so when the request hits the `:logger` middleware, the header is already there.
OTOH, when you use the recommended `http.request :token_auth`, you're injecting a middleware that will add the header on each request.
In order to see that header in your logs though, the `:logger` middleware needs to be inserted AFTER the `:token_auth` middleware.

I hope this makes sense, we're definitely getting close!

ybiquitous:
@iMacTia Thanks! I understand it. 👍🏼 

When I moved `http.response :logger` as below,

```diff
--- a/lib/octokit/connection.rb
+++ b/lib/octokit/connection.rb
@@ -115,6 +115,7 @@ module Octokit
         elsif application_authenticated?
           http.request :basic_auth, @client_id, @client_secret
         end
+        http.response :logger
       end
     end
 
```

I got the following output.

```
I, [2021-09-23T01:08:18.198989 #10907]  INFO -- request: GET https://api.github.com/
I, [2021-09-23T01:08:18.199023 #10907]  INFO -- request: Accept: "application/vnd.github.v3+json"
User-Agent: "Octokit Ruby Gem 4.21.0"
Content-Type: "application/json"
Authorization: "Token token=\"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx\""
I, [2021-09-23T01:08:18.206699 #10907]  INFO -- response: Status 401
```

The reason of the test failures seems to be a diff in the value of `Authorization`:

```diff
-Authorization: "token xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
+Authorization: "Token token=\"xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx\""
```

iMacTia:
Aha! Yes that's exactly how the `token_auth` works. I was surprised as well while refactoring that part the other day and I suspect this is coming from some super-old pre-Oauth implementation of auth tokens 😂
If `Authorization: "token xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"` is what you want, then you can use the plain authorisation middleware instead, like you do for Bearer tokens:

```ruby
http.request :authorization, 'token', @access_token
```

I know, this is super-confusing. And exactly the reason why I decided to remove it in Faraday 2.x 🤦‍♂️

ybiquitous:
Thanks! I've pushed c7b5fa2b to fix the token failure. 😄

ybiquitous:
https://github.com/octokit/octokit.rb/pull/1359#issuecomment-925084878

When following redirects, the `Authorization` header is not deleted... 🤔 

https://github.com/octokit/octokit.rb/blob/d729c0dfacf4aea854ac4899471bbbf59aa33035/lib/octokit/middleware/follow_redirects.rb#L89

iMacTia:
@ybiquitous so close! Checking the error it seems like this might also be a middleware order issue.

> Octokit::Client redirect handling drops authentication info when redirecting to a different host

The expectation is that following a redirect, the `Authorization` header should not be present, but that's not the case and the test fails.
Now, I believe this might be due to the fact that the `follow_redirect` middleware is inserted before the `authorisation` one(s).
More precisely, it's added [here](https://github.com/octokit/octokit.rb/blob/d1628ebd3e71aeb1534b3115c8bf37e8fb42068f/lib/octokit/default.rb#L29).

So the stack looks like this:

```
...
follow_redirects
authorisation (or basic_auth)
...
```

When a 3xx response is received, the `follow_redirects` middleware strips the `Authorisation` header before following, but the `authorisation` middleware after that will put it back!

If the two middleware were swapped, then this wouldn't happen anymore.

The reason this worked with the old implementation was that the header was previously added during the connection initialisation, rather than by a middleware during the request lifecycle

ybiquitous:
@iMacTia Thanks. As you thought, the `authorization` middleware inserts the `Authorization` header after the `follow_redirects` middleware deletes the header.

Is there a way to exclude the `authorization` middleware in the `follow_redirects` middleware?

iMacTia:
@ybiquitous the recommended way would be for the middleware to be in the right order, BUT there's another workaround.
The authentication middleware(s) will not touch the request if the Authorisation header is already set (see [this line](https://github.com/lostisland/faraday/blob/1.x/lib/faraday/request/authorization.rb#L56)).

That means that instead of removing the authorisation header completely, you could set it to some placeholder value.
Appreciate this is far from ideal, but there's no other way to leave the middleware where it is and just skip through it right now

ybiquitous:
@iMacTia Thank you very much for your kind advice! 😄 
I've added such a workaround via 23b4a5e4.

iMacTia:
My pleasure! I'm glad I could help ❤️
Feel free to mention me in future if you have any issue 🙌 !

hmharvey:
Thanks to everyone for your work and patience on this 💖

PR #1375, opened 2021-10-20
[fix] Escape `ref` in `archive_link`

max611:
When using `archive_link` with Unicode characters we have the following error: ` URI must be ascii only "repos/octokit/octokit.rb/tarball/\u{1F419}\u{1F431}"`

This PR fixes the issue with using `CGI.escape` but I am open to any other alternative, I saw that `CGI.escape` was already used in the codebase.

This is a breaking change since it's possible that someone was already using `CGI.escape` and we would escape it twice.


lib/octokit/client/contents.rb
```diff
@@ -156,7 +156,7 @@ def delete_contents(repo, path, message, sha, options = {})
       # @example Get archive link for octokit/octokit.rb
       #   Octokit.archive_link("octokit/octokit.rb")
       def archive_link(repo, options={})
-        repo_ref = options.delete :ref
+        repo_ref = CGI.escape(options.delete :ref)
```
nickfloyd, line 159: We recently merged in a change to remove many of the references to CGI in favor of ERB.  ERB actually URL encodes things closer to the defined spec like spaces to %20 instead of `+` like CGI does.  What are your thoughts / feels on using ERB instead?

```suggestion
        repo_ref = ERB::Util.url_encode(options.delete :ref)
```

nickfloyd (changes requested):
Hey @max611 thank you for putting in the work to make this change ❤️ .  I left a comment about using `ERB` vs `CGI` - let me know what you think about the suggestion and we'll try to get this merged in.  Thanks!

lib/octokit/client/contents.rb
max611: Makes sense to me!

nickfloyd (approved):

PR #1382, opened 2021-12-13
Correctly raise Octokit::TooManyRequests when hitting secondary rate limit

jasonopslevel:
The server returns 403 for a large swath of errors, and then client then
re-raises them as more appropriate errors via inspecting the body of the
response. Adding a case to raise Octokit::TooManyRequests for the secondary
rate limit error as well.

For example, the body returned for a 'primary' rate limit:

```
API rate limit exceeded for installation ID XXXXXXXX. // See: https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting
```

and the body for the secondary rate limit: 

```
You have exceeded a secondary rate limit. Please wait a few minutes before you try again. // See: https://docs.github.com/en/free-pro-team@latest/rest/overview/resources-in-the-rest-api#secondary-rate-limits
```


lib/octokit/error.rb
```diff
@@ -68,6 +68,8 @@ def self.error_for_401(headers)
     def self.error_for_403(body)
       if body =~ /rate limit exceeded/i
```
jasonopslevel, line 69: An alternative would be to instead just do:

```suggestion
      if body =~ /rate limit/i
```

However, went with the more explicit approach for now.

joelthecoder (approved):
Looks reasonable to me! I just encountered this same case, so thank you for putting in the work! 😄

PR #1413, opened 2022-04-20
Relax sawyer so we can use 0.9.x

BobbyMcWho:
We need this to resolve https://github.com/danger/danger/issues/1349


BobbyMcWho:
closing as #1411 resolves much more than just sawyer

PR #1414, opened 2022-04-27
replace git.io link

wonda-tea-coffee:
@see
https://github.blog/changelog/2022-04-25-git-io-deprecation/


nickfloyd (approved):
Thanks for the change @wonda-tea-coffee!  ✅  LGTM!

PR #1419, opened 2022-05-25
Ensures that any nil parameters being passed in will initalize with Octokit's defaults instead of nil

nickfloyd:
This replaces @akerl original [PR](https://github.com/octokit/octokit.rb/pull/1117) due to heavy deviation of base branches and the age of the changeset.  The source change was pulled directly from his PR and a test was added to ensure the change.

From #1117 

> This should fix https://github.com/octokit/octokit.rb/issues/1113 . The issue there affected 2 of my gems as well. It looks like the fix for https://github.com/octokit/octokit.rb/issues/1088 accidentally affected nil values as well as false. Because nil is a common pattern for pass-through configuration, parameters set to nil should still be overridden by module defaults.

### DIFF
--- a/spec/octokit/client_spec.rb
+++ b/spec/octokit/client_spec.rb
@@ -44,6 +44,12 @@ describe Octokit::Client do
         }
       end
 
+      it "defaults are used when parameters are nil" do
+        new_opts = @opts.merge(:api_endpoint => nil)
+        client = Octokit::Client.new(new_opts)
+        expect(client.api_endpoint).to eq(Octokit.api_endpoint)
+      end
+
       it "overrides module configuration" do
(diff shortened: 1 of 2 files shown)


spec/octokit/client_spec.rb
```diff
@@ -44,6 +44,12 @@
         }
       end
 
+      it "defaults are used when parameters are nil" do
```
mattpage, line 47:
```suggestion
      it "uses defaults when parameters are nil" do
```

mattpage: This just reads better from left to right

mattpage (approved):
One minor nit. Otherwise 👍

PR #1420, opened 2022-05-27
Allow autoloading Octokit::Repository

waiting-for-dev:
Hello!

Since dd5612e269336ecd43f7f12dbae077ecbef4cea2, it's not possible to reference `Octokit::Repository` when only `lib/octokit.rb` file has been required.

Steps to reproduce:

> require 'octokit'
> Octokit::Repository # => NameError

This is hitting as at Solidus, where our extensions try to directly reference `Octokit::Repository` as part of auto discovering the repository to generate a Changelog from. See https://github.com/solidusio/solidus_dev_support/blob/0764caad00f1b755d2cfdeef15762d85d5f58a01/lib/solidus_dev_support/rake_tasks.rb#L81. We found the problem since the last release yesterday.

Thanks for your work and, please, don't hesitate to ask us whatever you might need.

PR #1425, opened 2022-06-05
Require `octokit/rate_limit` from `lib/octokit/error.rb` so `RateLimit` is available

### LINKED ISSUE #1422: NameError: uninitialized constant Octokit::RateLimit
In 4.22 we had a unit test that raised this error, so we could test that our code behaved accordingly.  With the upgrade to 4.23, when we try to

```ruby
raise Octokit::TooManyRequests
```

we get the following:

```
NameError: uninitialized constant Octokit::RateLimit

     @context = Octokit::RateLimit.from_response(@response)
                       ^^^^^^^^^^^
from ~/.rbenv/versions/3.1.2/lib/ruby/gems/3.1.0/gems/octokit-4.23.0/lib/octokit/error.rb:39:in `build_error_context'
```

This can be reproduced easily with a new project that only depends on Octokit 4.23, and using `pry` to raise it.  I have put together a repository with that reproduction here:

https://github.com/collinsauve/Octokit-toomanyrequests-error

Additionally, the same issue *cannot* be reproduced in 4.22 as shown in this branch of that repository:

https://github.com/collinsauve/Octokit-toomanyrequests-error/tree/4.22

timrogers:
`Octokit::Error#build_error_context` uses `RateLimit.from_response` but does not require the file for `RateLimit` explicitly.

This can lead to cases where the `RateLimit` module is not available when needed, causing an error.

This requires the file explicitly where the module is used.

Fixes #1422.

### DIFF
--- a/lib/octokit/error.rb
+++ b/lib/octokit/error.rb
@@ -1,3 +1,5 @@
+require 'octokit/rate_limit'
+
 module Octokit
   # Custom error class for rescuing from all GitHub errors
   class Error < StandardError


timrogers:
The underlying issue here was introduced by #1351, which switched from loading all of Octokit when `require 'octokit'` is called to lazy-loading parts of the gem the first time `Octokit::Client` is referenced.

The end result is that if you instantiate an `Octokit::Error` subclass before you refer to `Octokit::Client`, it will fail because `Octokit::RateLimit` hasn't been loaded.

It's likely that there will be other similar cases, but it's hard to trace all of them because of the complex tree of how things are required. It seems reasonable just to fix this one case for now, rather than throw the baby out with the bathwater and revert the autoloading.

timrogers:
Looking at other recent PRs, #1420 is actually address the same problem, but affecting `Octokit::Repository` instead of error classes.

This makes me wonder if lazy-loading is just a bit problematic for users. 

__It means that anyone referring to a module inside Octokit directly has to manually require the relevant file in order to be sure that it's available)). This is all the more dangerous as you have to refer to a file path, effectively making internal file paths part of the SDK's public API.

At the same time, it would be a shame to lose lazy-loading, as it trims a fair bit of time from the boot-up of people's apps per #1351.

I wonder if we should offer users the choice of eager loading and lazy loading. One way to do that would be to add a `lib/octokit/all.rb` file which requires everything, and then people could do the following in their `Gemfile`:

```ruby
gem 'octokit', '4.23.0', require: 'octokit/all'
```

If we do this, the most interesting question is "what should be the recommended approach in the readme?". Should we suggest that people require everything and offer lazy-loading as a power user feature? Or should we recommend the lazy-loading approach and offer "requiring everything" is an advanced option for people delving into internals a bit more?

I prefer the former to make the experience of using Octokit as smooth as possible, especially given that referring to classes inside the gem is likely to be a common case - especially for errors which you `rescue`.

timrogers:
After further discussion with @nickfloyd, I think the right option is to revert #1420 for now as it is presenting itself as a breaking change in a few places. We can then work out a better, opt-in way to do lazy-loading.

timrogers:
Closing in favour of #1428.

PR #1428, opened 2022-06-06
Reverts autoloading of modules change

nickfloyd:
The original change introducing autoloading https://github.com/octokit/octokit.rb/pull/1351 was introduced in [v4.23.0](https://github.com/octokit/octokit.rb/releases/tag/v4.23.0).

For the most part, this change is/was beneficial but after @collinsauve highlighted an https://github.com/octokit/octokit.rb/issues/1422 that he was having as a result of this change we realized other areas where autoloading could cause issues due to modules not being loaded or load order even.

More details:
 https://github.com/octokit/octokit.rb/pull/1351 | https://github.com/octokit/octokit.rb/pull/1425 | https://github.com/octokit/octokit.rb/issues/1422


timrogers (approved):

nickfloyd:
Fixes: https://github.com/octokit/octokit.rb/pull/1428

nickfloyd:
Fixes: https://github.com/octokit/octokit.rb/issues/1422

PR #1431, opened 2022-06-07
Add support for lazy-loading by requiring `octokit/lazy`, rather than just `octokit`

timrogers:
#1351 introduced support for "lazy loading" in the library rather than loading everything upfront. @gmcgibbon ran some benchmarks and found that this approach shaved 400ms off their application startup.

This was a nice idea - however, it ended up unexpected being a breaking change because users were relying on all classes and modules in Octokit being loaded upfront (e.g. so `rescue Octokit::TooManyRequests` works). We ended up reverting the change in #1428.

This introduces new, opt-in lazy-loading. All you need to do to take advantage of it is `require 'octokit/lazy'` instead of requiring plan old `'octokit'`.

In our CI run, we test the gem with both methods of requiring, which should help to avoid any unexpected surprises when we release changes.

### DIFF
--- a/.github/workflows/octokit.yml
+++ b/.github/workflows/octokit.yml
@@ -56,3 +56,9 @@ jobs:
           GITHUB_CI: 1
           RUBYOPT: --enable-frozen-string-literal
         run: bundle exec rspec -w
+      - name: Test with RSpec in lazy-loaded mode
+        env:
+          GITHUB_CI: 1
+          RUBYOPT: --enable-frozen-string-literal
+          OCTOKIT_REQUIRE: octokit/lazy
+        run: bundle exec rspec -w
--- a/Gemfile
+++ b/Gemfile
@@ -42,4 +42,4 @@ group :test, :development do
   gem 'rubocop'
 end
 
-gemspec
+gemspec require: false
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ client.readme 'al3x/sovereign', :accept => 'application/vnd.github.html'
 
 ## Installation
 
-Install via Rubygems
+Install via RubyGems
 
     gem install octokit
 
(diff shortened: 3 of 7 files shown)


timrogers:
@gmcgibbon How do you feel about this approach? It allows you to opt-in for faster boot times, whilst not ending up being a breaking change in certain cases.

gmcgibbon (commented):
I think that we can still ship an autoloaded version of the gem without having to opt-in, it will just require a bigger refactor of each file. If we make it so each file only defines one module (and no peer modules like the error.rb file does by defining several error classes, there should be no problem).

The issue here is that that could take some time, and this approach will work for now, but the lazy loaded version of the client in this PR has the same bugs that will eventually need to be fixed.

Since I'm not the maintainer of this gem, it is not my decision, but I would refactor the file structure and then try re-introducing autoloading when this is done.

timrogers:
Thanks for taking a look at this so quickly ❤️

You're right - it would be possible to use `autoload` declarations throughout, in place of the current `require`s, and then everything can be lazy loaded seamlessly. I believe that can even work where one file provides multiple modules - we don't necessarily have to split them. 

However, even with that in mind, I'd still argue that it might be better for lazy-losing to be optional. Some users would likely prefer the work to happen when their app boots, rather than adding "warmup time" to their first request (as an example). 

What do you think? If we do want to provide the option of eager loading, then we could merge this and then, as a separate additive change, improve the lazy loading to be more reliable.

gmcgibbon:
> What do you think? If we do want to provide optional lazy loading, then we could merge this version and then improve the lazy loading process as a follow-up.

You could also just use Zeitwerk and support optional eager loading and autoloading like Rails does to be more flexible. However, I think it would be useful to look at benchmarks or profiles of the load time added by autoloading to a first request before deciding on anything.

timrogers:
I did consider Zeitwerk, but I was a little wary of adding runtime dependencies and I wasn't sure how it worked in the gem context. 

I'm happy to pause this PR, do some more digging and come back. 

I do think this solution is a good start, but it might be worth skipping straight to something better. 

Thank you for your feedback and for the original PR - I hope we can bring these performance improvements out soon.

BertsLifee (approved):

BertsLifee (approved):

PR #1432, opened 2022-06-07
Enable Rubocop in CI and automatically correct the offences we can correct

timrogers:


nickfloyd (approved):
❤️ for these changes and getting this all knocked out!

lib/octokit/client/contents.rb
```diff
@@ -78,17 +76,20 @@ def create_contents(*args)
...
           end
         end
-        raise ArgumentError.new("content or :file option required") if content.nil?
-        options[:content] = Base64.respond_to?(:strict_encode64) ?
-          Base64.strict_encode64(content) :
-          Base64.encode64(content).delete("\n") # Ruby 1.9.2
+        raise ArgumentError, "content or :file option required" if content.nil?
+
+        options[:content] = if Base64.respond_to?(:strict_encode64)
+  Base64.strict_encode64(content)
```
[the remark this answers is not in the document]
timrogers: It also looks like this could be removed entirely as it's targeting Ruby 1.9.2. That means that we can merge this as is, and come back in another PR to cut out this condition.

lib/octokit/client/pull_requests.rb
```diff
@@ -205,15 +203,15 @@ def pull_request_comment(repo, comment_id, options = {})
       #     "2d3201e4440903d8b04a5487842053ca4883e5f0", "lib/octokit/request.rb", 47)
       def create_pull_request_comment(repo, pull_id, body, commit_id, path, position, options = {})
         options.merge!({
-          :body => body,
-          :commit_id => commit_id,
-          :path => path,
-          :position => position
-        })
+                         body: body,
```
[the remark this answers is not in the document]
timrogers: There's another example later in the file as well.

lib/octokit/client/repositories.rb
```diff
@@ -159,19 +155,17 @@ def fork(repo, options = {})
       def create_repository(name, options = {})
         opts = options.dup
         organization = opts.delete :organization
-        opts.merge! :name => name
-        if opts.include? :is_template
-          opts = ensure_api_media_type(:template_repositories, opts)
-        end
+        opts.merge! name: name
```
timrogers, line 158: I'm also surprises by the lack of brackets here.

lib/octokit/client/repositories.rb
```diff
@@ -210,19 +204,19 @@ def transfer_repository(repo, new_owner, options = {})
       # @option options [Boolean] :include_all_branches `true` copies all branches from the template repository, `false` (default) makes it only copy the master branch.
       # @return [Sawyer::Resource] Repository info for the new repository
       def create_repository_from_template(repo, name, options = {})
-        options.merge! :name => name
+        options.merge! name: name
```
timrogers, line 207: Let's add brackets here as well

lib/octokit/client/service_status.rb
```diff
@@ -1,16 +1,14 @@
...
     #
     # @see https://status.github.com/api
     module ServiceStatus
-
       # Root for status API
       # @private
-      SUMMARY_ROOT    = 'https://www.githubstatus.com/api/v2/summary.json'
-      STATUS_ROOT     = 'https://www.githubstatus.com/api/v2/status.json'
-      COMPONENTS_ROOT = 'https://www.githubstatus.com/api/v2/components.json'
+      SUMMARY_ROOT    = 'https://www.githubstatus.com/api/v2/summary.json'.freeze
```
timrogers, line 9: We might as well add the frozen string literal declaration to each of our files, rather than do this. We test with frozen strings enabled already, so it shouldn't matter.

lib/octokit/client/users.rb
```diff
@@ -309,9 +307,9 @@ def emails(options = {})
       # @see https://developer.github.com/v3/users/emails/#add-email-addresses
       # @example
       #   @client.add_email('new_email@user.com')
-      def add_email(email, options = {})
+      def add_email(email, _options = {})
```
timrogers, line 310: It's pretty weird that this (and other) methods take this `options` argument but don't use it at all. I wonder if it's intentional. (It would be a breaking change to remove them.)

lib/octokit/configurable.rb
```diff
@@ -113,24 +117,22 @@ def same_options?(opts)
...
     # Base URL for generated web URLs
     #
     # @return [String] Default: https://github.com/
     def web_endpoint
-      File.join(@web_endpoint, "")
+      File.join(@web_endpoint, '')
     end
 
     def login
-      @login ||= begin
```
timrogers, line 131: I'm not sure what the significance of the `begin` is here, but Rubocop seems to want to remove it. This might take some research.

lib/octokit/default.rb
```diff
@@ -43,11 +41,10 @@ module Default
     end
 
     class << self
-
       # Configuration options
       # @return [Hash]
       def options
-        Hash[Octokit::Configurable.keys.map{|key| [key, send(key)]}]
+        Hash[Octokit::Configurable.keys.map {|key| [key, send(key)]}]
```
timrogers, line 47: This could use `Array#to_h`, which Rubocop automatically introduced in a similar place above. I'm not sure why it didn't do it here.

lib/octokit/error.rb
```diff
@@ -225,10 +223,10 @@ class Unauthorized < ClientError; end
   # Raised when GitHub returns a 401 HTTP status code
   # and headers include "X-GitHub-OTP"
   class OneTimePasswordRequired < ClientError
```
timrogers, line 225: I don't think this code - or anything else around OTPs - is relevant with modern GitHub authentication methods.

lib/octokit/rate_limit.rb
```diff
@@ -12,15 +11,14 @@ module Octokit
   #   @return [Integer] Number of seconds when rate limit resets
   #
   # @see https://developer.github.com/v3/#rate-limiting
-  class RateLimit < Struct.new(:limit, :remaining, :resets_at, :resets_in)
-
+RateLimit = Struct.new(:limit, :remaining, :resets_at, :resets_in) do
```
timrogers, line 14: Weird indentation here, or just the diff looking weird on GitHub?

lib/octokit/response/feed_parser.rb
```diff
@@ -1,19 +1,15 @@
 require 'octokit/response/base_middleware'
 
 module Octokit
-
   module Response
-
     # Parses RSS and Atom feed responses.
     class FeedParser < BaseMiddleware
```
timrogers, line 6: Do we have RSS and Atom feed responses anymore? 🤔 If not, this can go.

octokit.gemspec
```diff
@@ -1,23 +1,23 @@
...
   spec.homepage = 'https://github.com/octokit/octokit.rb'
   spec.licenses = ['MIT']
   spec.name = 'octokit'
   spec.require_paths = ['lib']
   spec.required_ruby_version = '>= 2.0.0'
   spec.required_rubygems_version = '>= 1.3.5'
-  spec.summary = "Ruby toolkit for working with the GitHub API"
+  spec.summary = 'Ruby toolkit for working with the GitHub API'
   spec.version = Octokit::VERSION.dup
+  spec.metadata['rubygems_mfa_required'] = 'true'
```
timrogers, line 22: We should ✂️ this, and then introduce it separately. It's a bit of a weird one to sneak in through Rubocop!

spec/helper.rb
```diff
@@ -10,12 +10,10 @@
 require 'base64'
 require 'jwt'
 # latest version of pry-byebug is not compatible with Ruby 3.2.0
-if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('3.2.0')
-  require 'pry-byebug'
-end
+require 'pry-byebug' if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('3.2.0')
```
timrogers, line 13: We should check if `pry-byebug` still has this problem and remove it when it doesn't

lib/octokit/client/contents.rb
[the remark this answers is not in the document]
timrogers: Issue created: https://github.com/octokit/octokit.rb/issues/1433

lib/octokit/client/pull_requests.rb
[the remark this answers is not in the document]
timrogers: It's fine - just weird in the diff.

lib/octokit/client/repositories.rb
timrogers:
```suggestion
        opts.merge!(name: name)
```

lib/octokit/client/repositories.rb
timrogers:
```suggestion
        options.merge!(name: name)
```

lib/octokit/client/service_status.rb
timrogers: Issue created: https://github.com/octokit/octokit.rb/issues/1435

lib/octokit/client/users.rb
```diff
@@ -45,14 +43,14 @@ def user(user=nil, options = {})
...
         options = options.merge({
-          :code => code,
-          :client_id => app_id,
-          :client_secret => app_secret,
-          :headers => {
-            :content_type => 'application/json',
-            :accept       => 'application/json'
-          }
-        })
+                                  code: code,
```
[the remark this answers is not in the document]
timrogers: All good!

lib/octokit/configurable.rb
timrogers: Looks like this is here for the reasons described [here](https://stackoverflow.com/questions/13279217/are-there-unintended-consequences-of-rubys-begin-end-without-rescue-use). Using brackets will achieve the same thing in a more idiomatic way.

spec/helper.rb
timrogers: Issue created: https://github.com/octokit/octokit.rb/issues/1439

timrogers:
Closing in favour of #1441

PR #1441, opened 2022-06-07
Enable Rubocop - the sequel

timrogers:
This PR is a follow up to #1432, again trying to enable Rubocop in CI and autocorrect what we can, whilst keeping the gem actually working 😅 ❤️ 

I decided to start again with a fresh PR because the original PR had a very ugly history - lots of embarrassing fiddling around getting the right `TargetRubyVersion` for Rubocop! - and I didn't want to destroy it as there were some useful comments in there.

It's worth noting that, to support linting compatible with Ruby 2.3 (which is the current version we actually support, despite what the gemspec claims!), we have to downgrade Rubocop to an earlier version.


timrogers:
Theoretically, Rubocop autocorrection shouldn't lead to any substantive changes in the meaning of your code, but it can make mistakes from time to time. I have a high level of confidence though that we're safe: the tests pass, and I have manually read through the changes to check for anything evident (although this is obviously challenging given the length of the diff).

PR #1443, opened 2022-06-08
Require multi-factor authentication to push new releases to RubyGems

### LINKED ISSUE #1438: Update gemspec to require multi-factor authentication to push releases
We should ✂️ this, and then introduce it separately. It's a bit of a weird one to sneak in through Rubocop!

_Originally posted by @timrogers in https://github.com/octokit/octokit.rb/pull/1432#discussion_r891338223_

timrogers:
This updates our gemspec to opt in to enforce multi-factor authentication (MFA) whenever we want to push releases of the gem, yank releases or add or remove owners.

This will increase the security of our users by making it more difficult for a bad actor to release a version of Octokit.rb.

After this is merged, the change will take effect from the first new release.

Fixes #1438.


nickfloyd (approved):
❤️ 🔏

PR #1446, opened 2022-06-13
Adds details on how to run a manual file integrity check

nickfloyd:
Adds details on how to check the files that are packed in the gem to help anyone who intends on releasing this make sure they are building and releasing this as intended.

This adds a script that is executed when the release or package scripts are run to naively check the permissions of the files in the gem - just as a safeguard.

Additionally, this adds parameters to script/release so that dry runs can be performed 

Hopefully, we'll be able to automate this, but for now, if anyone tries to release from something like a code space, for instance, then the gem will be released with world writeable files. We need to make sure that this is verified before release.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -2,16 +2,47 @@
 
 1. Create a list of all the changes since the prior release
     1. Compare the latest release to master using https://github.com/octokit/octokit.rb/compare/`${latest}`...master
-    1. Open the linked pull requests from all the `Merge pull request #...` commits
-    1. For all non-documentation PRs, copy title (including pull request number) into markdown list items
-    1. (optional, but nice) Sort into logical buckets, like "support for additional endpoints", "enhancements", "bugfixes"
-    1. Reorganize to put the pull request number at the start of the line
-1. Ensure there are no breaking changes _(if there are breaking changes you'll need to create a release branch without those changes or bump the major version)_
-1. Update the version
+    2. Open the linked pull requests from all the `Merge pull request #...` commits
+    3. For all non-documentation PRs, copy title (including pull request number) into markdown list items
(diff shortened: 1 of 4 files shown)


RELEASE.md
```diff
@@ -2,16 +2,47 @@
...
+    3. For all non-documentation PRs, copy title (including pull request number) into markdown list items
+    4. (optional, but nice) Sort into logical buckets, like "support for additional endpoints", "enhancements", "bugfixes"
+    5. Reorganize to put the pull request number at the start of the line
+2. Ensure there are no breaking changes _(if there are breaking changes you'll need to create a release branch without those changes or bump the major version)_
+3. Update the version
     1. Update the constant in `lib/octokit/version.rb`
-    1. Commit and push directly to master
-1. Run the `script/release` script to cut a release
-1. Draft a new release at https://github.com/octokit/octokit.rb/releases/new containing the curated changelog
+    2. Run the "File integrity check"
```
timrogers, line 12: It feels like this would make more sense as a separate step, rather than a sub-step of updating the version. What do you think of doing it directly before running `script/release`?

RELEASE.md
```diff
@@ -2,16 +2,47 @@
...
+From the root of octokit.rb
+
+```
+> gem build *.gemspec
+```
+
+Use the version from the build in the next commands
+
+```
+> tar -x -f octokit-#.##.#.gem
```
timrogers, line 36: It'd be really nice to have a more automated way to do these checks recursively down the directory tree - but not essential for now.

RELEASE.md
nickfloyd: Sure that sounds good.

nickfloyd:
Fixes: https://github.com/github/public-interfaces-sdk/issues/44

PR #1452, opened 2022-07-11
Trigger GitHub Actions workflows on the `main` branch

timrogers:
Currently, the default branch for this repo is `4-stable` (not `main` or `master`) and we have some old code which was proposed to be the v5.x version of Octokit.rb in the `5-alpha` branch.

We now want to release a new major version of Octokit to drop support for old Ruby versions - see https://github.com/octokit/octokit.rb/pull/1424.

To do that, we'll delete the unused `master` branch and rename `4-stable` to `main`.

Ahead of that change, this updates our GitHub Actions workflows to support the name `main`. Once the rename is done, we can remove all of the references to the old branch names.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -13,10 +13,14 @@ name: "CodeQL"
 
 on:
   push:
-    branches: [ 4-stable ]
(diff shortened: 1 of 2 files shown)


nickfloyd (approved):
LGTM!

PR #1454, opened 2022-07-11
Drop support for EOL'd Ruby versions 2.6 and earlier

timrogers:
This drops support for older Ruby versions which have reached end-of-life, so Octokit.rb will only work with Ruby 2.7, 3.0 and 3.1.

Today, the gemspec claims to support v2.0 onwards, but in practice, only v2.3 onwards are tested in CI and actually working.

### DIFF
--- a/.github/workflows/octokit.yml
+++ b/.github/workflows/octokit.yml
@@ -26,18 +26,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu ]
-        ruby: [ 2.3, 2.4, 2.5, 2.6, 2.7, '3.0', '3.1', head ]
+        ruby: [ 2.7, '3.0', '3.1', head ]
         faraday: [ '~> 1.0', '~> 2.0' ]
-        exclude:
-          # Faraday 2.x only works with Ruby versions 2.6 onwards.
-          # Fortunately, the gem supports Faraday 1.x, so we can just
-          # test against that.
-          - ruby: 2.5
-            faraday: '~> 2.0'
-          - ruby: 2.4
-            faraday: '~> 2.0'
-          - ruby: 2.3
-            faraday: '~> 2.0'
     env:
       FARADAY_VERSION: ${{ matrix.faraday }}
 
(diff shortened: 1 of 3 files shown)


timrogers:
@nickfloyd The missing checks are okay as those versions are now dropped from the test matrix 😊

.github/workflows/octokit.yml
```diff
@@ -26,18 +26,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [ ubuntu ]
-        ruby: [ 2.3, 2.4, 2.5, 2.6, 2.7, '3.0', '3.1', head ]
+        ruby: [ 2.7, '3.0', '3.1', head ]
         faraday: [ '~> 1.0', '~> 2.0' ]
```
nickfloyd, line 30: I am pretty sure we can get rid of the 1.0 reference here... Let me know if that sounds right.

```suggestion
        faraday: '~> 2.0'
```

timrogers: We can still continue to support Faraday 1.x. Dropping support for Faraday 1.x is often pretty painful as there are lots of packages out there that still require it.

nickfloyd (approved):
❤️

Mdrajjak334280:
Den-bkash

.github/workflows/octokit.yml
nathaliepoulin: > Den-bkash

debug_concurrency_issues_tutorial_procedure_4

Djstiers110987 (approved):

Djstiers110987 (approved):

.github/workflows/octokit.yml
yehiamitro111: > I am pretty sure we can get rid of the 1.0 reference here... Let me know if that sounds right.
> 
> ```suggestion
>         faraday: '~> 2.0'
> ```

yehiamitro111 (commented):
Thank you

Rucellmai61 (changes requested):
[]()

afriadishariff-lab:
> This drops support for older Ruby versions which have reached end-of-life, so Octokit.rb will only work with Ruby 2.7, 3.0 and 3.1.
> 
> Today, the gemspec claims to support v2.0 onwards, but in practice, only v2.3 onwards are tested in CI and actually working.

PR #1461, opened 2022-07-13
Add GitHub Actions to Dependabot

petergoldstein:
This PR adds Dependabot for GitHub Actions as described here - https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

As the checkout action is currently out of date, Dependabot will open a PR to update it if and when this PR is merged.


timrogers (approved):
Nice idea! I forgot this. Thanks for the PR.

Issue #1447, opened 2022-06-17
Remove all previews logic from the SDK


nickfloyd:
REST API previews have now either been [promoted or deprecated](https://github.blog/changelog/2021-10-14-rest-api-preview-promotions/) from the GitHub REST API.

They were removed or promoted and the preview mechanism is no longer supported in [api.github.com](https://github.com/github/rest-api-description/tree/main/descriptions/api.github.com) and [ghes-3.3](https://github.com/github/rest-api-description/tree/main/descriptions/ghes-3.3) and above for the REST API.  

The logic should be removed from this SDK.

PR #1464, opened 2022-07-25
Remove REST API previews logic from the SDK now that all previews have been promoted 

nickfloyd:
Fixes: https://github.com/octokit/octokit.rb/issues/1447

This change set removes all preview logic from this SDK.  All REST API previews we promoted last year and the logic to enforce/use previews is obsolete.

NOTE: I'll do a follow-up when time allows regenerating the cassettes to remove the headers from those files completely

Apologies for the large change set - previews were scattered across much of the SDK.


.rubocop.yml
```diff
@@ -4,8 +4,7 @@ AllCops:
   TargetRubyVersion: 2.3
   NewCops: enable
 Metrics/BlockLength:
-  Exclude:
-    - spec/**/*_spec.rb
+  IgnoredMethods: ['describe', 'context']
```
nickfloyd, line 7: Reworked this to filter blocks on test methods instead of whole files due to helpers and the like being in those files and should be crafted well.

spec/octokit/client/repositories_spec.rb
```diff
@@ -62,23 +62,6 @@
     end
   end
 
-  describe '.edit_repository', :vcr do
```
nickfloyd, line 65: Simplified and moved this test down into the `with repository` group - there was also a new cassette as a result of this change.

spec/octokit/client/organizations_spec.rb
```diff
@@ -332,22 +331,10 @@
...
       @client.organization_membership(
         test_github_org,
-        user: test_github_login,
-        accept: 'application/vnd.github.moondragon+json'
+        user: test_github_login
       )
       assert_requested :get, github_url("/orgs/#{test_github_org}/memberships/#{test_github_login}")
     end
-
-    it 'returns an organization membership for a given user by the orgs id' do
```
nickfloyd, line 341: This is no longer a valid test due to the REST API (OpenAPI descriptions) no longer allow ids to be passed in the query string for orgs

tarebyte (approved):

PR #1472, opened 2022-08-22
Adds ability to update pull request branch

### LINKED ISSUE #1471: Update pull request branch missing?
:wave:

Looking through the REST API docs I noticed that pull requests have a [`update-branch`](https://docs.github.com/en/rest/pulls/pulls#update-a-pull-request-branch) endpoint that will update the pull request branch with the latest upstream changes from the base branch but that this functionality is missing from the Octokit Ruby client. Is this intentional or just an oversight? Would a contribution to add this functionality be welcomed?

bjedrocha:
Resolves #1471 

As per https://docs.github.com/en/rest/pulls/pulls#update-a-pull-request-branch. I couldn't get the specs to work with VCR despite creating a test repo and setting all relevant ENV vars. I decided to stub the request instead.

### DIFF
--- a/lib/octokit/client/pull_requests.rb
+++ b/lib/octokit/client/pull_requests.rb
@@ -282,9 +282,9 @@ module Octokit
       # @param repo [Integer, String, Hash, Repository] A GitHub repository
       # @param number [Integer] Number of pull request
       # @param options [Hash] Optional parameters (e.g. expected_head_sha)
-      # @return [Sawyer::Resource] Info regarding the update if successful
+      # @return [Boolean] True if the pull request branch has been updated
       def update_pull_request_branch(repo, number, options = {})
-        put "#{Repository.path repo}/pulls/#{number}/update-branch", options
+        boolean_from_response(:put, "#{Repository.path repo}/pulls/#{number}/update-branch", options)
       end
 
       # Merge a pull request
--- a/lib/octokit/client/pull_requests.rb
+++ b/lib/octokit/client/pull_requests.rb
@@ -276,6 +276,17 @@ module Octokit
       end
       alias pull_files pull_request_files
 
+      # Update a pull request branch
+      #
+      # @see https://developer.github.com/v3/pulls/#update-a-pull-request-branch
+      # @param repo [Integer, String, Hash, Repository] A GitHub repository
+      # @param number [Integer] Number of pull request
+      # @param options [Hash] Optional parameters (e.g. expected_head_sha)
+      # @return [Sawyer::Resource] Info regarding the update if successful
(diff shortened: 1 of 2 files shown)


lib/octokit/client/pull_requests.rb
```diff
@@ -276,6 +276,17 @@ def pull_request_files(repo, number, options = {})
...
 
+      # Update a pull request branch
+      #
+      # @see https://developer.github.com/v3/pulls/#update-a-pull-request-branch
+      # @param repo [Integer, String, Hash, Repository] A GitHub repository
+      # @param number [Integer] Number of pull request
+      # @param options [Hash] Optional parameters (e.g. expected_head_sha)
+      # @return [Sawyer::Resource] Info regarding the update if successful
+      def update_pull_request_branch(repo, number, options = {})
+        put "#{Repository.path repo}/pulls/#{number}/update-branch", options
```
timrogers, line 287: What do you thinking of using `#boolean_from_response` here? Looking at the API, it appears that the response is not super useful - you just want to know if it worked or not.

```suggestion
        boolean_from_response(:put, "#{Repository.path repo}/pulls/#{number}/update-branch", options)
```

spec/octokit/client/pull_requests_spec.rb
```diff
@@ -136,6 +136,15 @@
     end
   end # new @pull methods
 
+  # stub this so we don't have to set up new fixture data
+  describe '.update_pull_request_branch' do
```
timrogers, line 140: We should test the return value of `#update_pull_request_branch` in here.

timrogers (changes requested):
👋🏻 Thanks for taking the time to make this PR! I have a suggestion on how we could make the behaviour of the method a little nicer, and on how we test it.

bjedrocha:
Thanks for the quick review - I can address that feedback shortly. What are your thoughts on having the API called stubbed vs. having a cassette for it?

timrogers:
> Thanks for the quick review - I can address that feedback shortly. What are your thoughts on having the API called stubbed vs. having a cassette for it?

I'm okay for it to be stubbed. Could we stub it with an actual response for better test. overage?

bjedrocha:
Addressed your feedback. I've changed `update_pull_request_branch` to return a simple boolean if the request succeeds. I also improved the test coverage by adding specs for the requests that fail. I didn't stub the response body as we're not doing anything with anyway but that can always be added if you think it provides value.

tride1997:
1472

Vào 23:32, Th 3, 23 thg 8, 2022 Bart Jedrocha ***@***.***> đã
viết:

timrogers (approved):

PR #1475, opened 2022-08-30
Tweak release documentation and process

timrogers:
This PR makes a number of improvements to our documentation and process for releasing a new version of the gem, contained in `RELEASE.md` and `script/release`. Specifically, it:

* fixes the instructions so you don't end up with two release commits - one created manually by following the instructions, and one created by `script/release`
* updates `script/release` to use the version number of the latest version when generating the commit message and Git tag, not the _oldest_ version (this is fine if you only have one `.gem` file, but not if you have old ones hanging around!)
* tweaks the commit message from the release script to be a simpler `v1.2.3` not `Release 1.2.3`


nickfloyd:
> * updates `script/release` to use the version number of the latest version when generating the commit message and Git tag, not the _oldest_ version (this is fine if you only have one `.gem` file, but not if you have old ones hanging around!)

Nice catch on this one ❤️

nickfloyd (approved):
LGTM!

PR #1476, opened 2022-09-02
Update "See Also" link for projects API client

michaelroudnitski:
The current link 404s, I believe the docs were moved to https://docs.github.com/en/rest/projects
Screen Shot 2022-09-02 at 2 16 31 PM


timrogers (approved):
Thanks for spotting this and fixing it!

PR #1483, opened 2022-09-14
Correct documentation URL for `#artifact_download_url`

timrogers:
With each client method, we include its URL in our API documentation on https://docs.github.com. After releasing v5.6.0, I spotted that the URL for `#artifact_download_url` was incorrect, pointing to a different API. This fixes it.


timrogers:
FYI @lerebear

Issue #1482, opened 2022-09-13
TooLargeContent Exception no longer thrown


goldstar611:
I'm not sure how long this has regressed but our github enterprise (version 3.6.1) instance is delivering the following message inside a Sawyer::Resource

```
Either the raw or object media type must be used for this endpoint for blobs between 1-100 MB. The raw type will return the file contents, and the JSON type will return the file's metadata. Please try again with a valid media type.
```

We can see there's a regex in error.rb#L80 that controls when an exception is raised:
https://github.com/octokit/octokit.rb/blob/15be7251e795044f2bb9537db7c58c95fdd4cbec/lib/octokit/error.rb#L80

This message does not match the regex because the text has changed from
"blobs up to XXX MB"
to
"blobs between X-YYY MB."

The TooLargeContent exception and regex was introduced in PR #1106

goldstar611:
I'm happy to PR this issue with either a non-greedy match like 
`/returns blobs .*? MB/i` 
or update the regex for a stricter match like
`/returns blobs (up to|between) [0-9\-]+ MB/i`

timrogers:
Great spot on this! ✨ It'd be great to have a PR to fix this. I'd go for the stricter regex I think to avoid any surprises.

PR #1484, opened 2022-09-14
Update regex for Octokit::TooLargeContent

goldstar611:
The error message changed at some point and Octokit no longer throws Octokit::TooLargeContent
- Github Enterprise 3.6.1 is affected
- Fixes https://github.com/octokit/octokit.rb/issues/1482

In Github Enterprise 3.6.1 (and probably others), the message returned in the Sawyer::Resource is 

```
Either the raw or object media type must be used for this endpoint for blobs between 1-100 MB. The raw type will return the file contents, and the JSON type will return the file's metadata. Please try again with a valid media type.
```

### DIFF
--- a/lib/octokit/error.rb
+++ b/lib/octokit/error.rb
@@ -77,7 +77,7 @@ module Octokit
         Octokit::TooManyRequests
       when /login attempts exceeded/i
         Octokit::TooManyLoginAttempts
-      when /returns blobs up to [0-9]+ MB/i
+      when /returns blobs (up to|between) [0-9\-]+ MB/i
         Octokit::TooLargeContent
       when /abuse/i
         Octokit::AbuseDetected


lib/octokit/error.rb
```diff
@@ -77,7 +77,7 @@ def self.error_for_403(body)
         Octokit::TooManyRequests
       when /login attempts exceeded/i
         Octokit::TooManyLoginAttempts
-      when /returns blobs up to [0-9]+ MB/i
+      when /returns blobs (up to|between) [0-9\-]+ MB/i
```
timrogers, line 80: Would you be able to add a test covering the new error message? We have a test for the old error message in `spec/octokit/client_spec.rb:871` which can be pretty much copy-and-pasted.

timrogers (changes requested):
Thanks for taking the time to fix this! It'd be great to add test coverage for this so we can be sure that it works.

lib/octokit/error.rb
goldstar611: Working on that now, and the regex needs a small modification as well. :)

goldstar611: Sorry for the commit spam. I added a test and double checked with regexr.com so I think this is ready to go now. 

Also let me know if you'd rather use a non-capturing group for performance, I went with readability here.

timrogers (approved):

AmandaP87 (commented):
goldstar611:patch-1

PR #1488, opened 2022-09-23
Fix broken pagination in ActionsWorkflowRuns

luke-engle:
Pagination was broken due to the data in the API response containing a hash instead of an array. This change adds a block to concatenate the `workflow_runs` array within the hash of each page's data


luke-engle:
I believe there are still other methods that have pagination broken in the same way:
- [actions_secrets.rb](https://github.com/octokit/octokit.rb/blob/main/lib/octokit/client/actions_secrets.rb)
- [actions_artifacts.rb](https://github.com/octokit/octokit.rb/blob/main/lib/octokit/client/actions_artifacts.rb)
- potentially others?

Essentially anything that paginates and has a response that returns a hash instead of an array.

I don't have the time to fix and create tests for the others right away but I'll add it to my todo list and try to fix them within the next couple days (I may be able to find time tomorrow to get this done since it shouldn't take too long). I'm going to convert this PR to a draft until I address those issues as well that way we don't need to create multiple releases to address this

timrogers:
Thanks for picking this up ❤️ As you say, it would be great to fix a bunch of these in one PR if we can. 

This is making me wonder if we need to do something in the OpenAPI spec itself to describe how data is paginated.

luke-engle:
> Thanks for picking this up ❤️ As you say, it would be great to fix a bunch of these in one PR if we can.
> 
> This is making me wonder if we need to do something in the OpenAPI spec itself to describe how data is paginated.

Ya it seems like the Actions team (who came from Microsoft) are the ones who implemented the the 'total count', which is why pagination is broken in the first place. If it were up to me, I would say to make all the APIs consistent in what they return. In other words, don't return the total count, just return the data, then we wouldn't need to do any extra work to paginate.

But regardless, the work here is done! Sorry for the delay. I had a bit of a health episode last week that kept me out pretty much the whole week 😭

luke-engle:
@timrogers Who could I get to review this change, and when can we get a new release out?

timrogers (approved):

timrogers:
@luke-engle This looks good to me. Would you be able to re-push to see if we can get Actions to run the workflows?

nickfloyd:
@timrogers  & @luke-engle I've unblocked the workflows - it was waiting for workflow approval... I'll keep an eye on it to ensure all checks are running as expected.

nickfloyd:
Looks like there is a debug statement in the [tests](https://github.com/luke-engle/octokit.rb/blob/main/spec/octokit/client/actions_workflow_jobs_spec.rb#L93) that is causing a [problem](https://github.com/octokit/octokit.rb/actions/runs/3163265465/jobs/5269489898). 

If that gets removed, I think that the workflows should be able to run complete again.

luke-engle:
> Looks like there is a debug statement in the [tests](https://github.com/luke-engle/octokit.rb/blob/main/spec/octokit/client/actions_workflow_jobs_spec.rb#L93) that is causing a [problem](https://github.com/octokit/octokit.rb/actions/runs/3163265465/jobs/5269489898).
> 
> If that gets removed, I think that the workflows should be able to run complete again.

Thanks Nick! That debug statement has been removed 👍

timrogers:
@luke-engle Looks like we still have a linting issue - you should be able to clean it up automatically with `bundle exec rubocop -A`.

PR #1494, opened 2022-10-17
Remove Octokit::Client::Authorizations

### LINKED ISSUE #1429: Remove `Octokit::Client::Authorizations`, which targets the now-gone OAuth Authorizations API
We can remove this module because the APIs it exposes were [deprecated](https://developer.github.com/changes/2020-02-14-deprecating-oauth-auth-endpoint/) back in 2020. I can see in GitHub's logs that these APIs are indeed not being hit anymore 😅 

This should be done in a breaking version since it removes code that people could be depending on - even if technically it hasn't been working for the past 18 months anyway.

kfcampbell:
Fixes #1429. 

This PR should land in a breaking change as Tim noted in the above issue, even if the underlying APIs are no longer functional.


lib/octokit/client/authorizations.rb
```diff
@@ -1,185 +0,0 @@
...
-        boolean_from_response :delete, "authorizations/#{number}", options
-      end
-
-      # Check scopes for a token
-      #
-      # @param token [String] GitHub OAuth token
-      # @param options [Hash] Header params for request
-      # @return [Array<String>] OAuth scopes
-      # @see https://developer.github.com/v3/oauth/#scopes
-      def scopes(token = @access_token, options = {})
```
timrogers, line 130: It might be worth keeping this one? 🤔 It isn't specifically related to the Authorizations API and it could be useful for users. I'm not sure we have any replacement to it either, as I don't think there is a way to access headers in the SDK.

kfcampbell: Good point! Would you rather I keep the authorizations module around for this use case, or refactor this functionality somewhere else?

timrogers: Kinda up to you, honestly. I would probably move it ("authorizations" isn't really a sensible name, since I think it works for any token?).

nickfloyd (approved):

kfcampbell:
@timrogers Apologies for the force-push! It took me a minute to understand that vcr was reaching out to an actual endpoint and auto-generating failing cassettes when I first renamed the Authorizations module. 

Is this fix something like you had in mind? I've gone for `module Tokens` as a name, and included the `create_authorization` method as well since that's required to test `scopes`. Thoughts?

lib/octokit/client/tokens.rb
```diff
@@ -0,0 +1,79 @@
...
+      # @see https://developer.github.com/v3/oauth/#scopes Available scopes
+      # @see https://developer.github.com/v3/oauth_authorizations/#create-a-new-authorization
+      # @see https://developer.github.com/v3/oauth_authorizations/#get-or-create-an-authorization-for-a-specific-app
+      # @example Create a new authorization for user ctshryock's project Zoidberg
+      #  client = Octokit::Client.new(:login => 'ctshryock', :password => 'secret')
+      #  client.create_authorization({:scopes => ["public_repo", "gist"], :note => "Why not Zoidberg?", :note_url=> "https://en.wikipedia.org/wiki/Zoidberg"})
+      # @example Create a new OR return an existing authorization to be used by a specific client for user ctshryock's project Zoidberg
+      #  client = Octokit::Client.new(:login => 'ctshryock', :password => 'secret')
+      #  client.create_authorization({:idempotent => true, :client_id => 'xxxx', :client_secret => 'yyyy', :scopes => ["user"]})
+      def create_authorization(options = {})
```
timrogers, line 32: Why is this required to test `#scopes`? That method looks like it should work with any token and get its permissions.

timrogers: Could we not just call `#scopes` directly with a token, rather than creating one with `#create_authorization`?

kfcampbell: Welllllll of course we could do that! Forgive my slowness, and please take another look!

kfcampbell:
I'm unsure of the proper Ruby practice around cassettes or how to autogenerate/remove them. The create_authorizations cassettes are now safe to remove so I've done so manually.

kfcampbell:
@timrogers Do you think this is okay to merge after your update?

PR #1495, opened 2022-10-17
Remove Octokit::Client::Hooks#available_hooks

### LINKED ISSUE #1430: Remove `Octokit::Client::Hooks#available_hooks`, built for the now-deprecated GitHub Services
`Octokit::Client::Hooks#available_hooks` was introduced to the SDK 9 years ago. At the time, it listed available Service Hooks for GitHub Services, but this was [deprecated](https://docs.github.com/en/developers/overview/replacing-github-services)  about 4 years ago.

We can remove this method, since it won't actually work, but it will be a breaking change since a method that existed will no longer exist.

kfcampbell:
Fixes #1430. 

As noted by Tim in the above issue, this should land in a breaking change even though the underlying API hasn't been operational for some time.

### DIFF
--- a/lib/octokit/client/hooks.rb
+++ b/lib/octokit/client/hooks.rb
@@ -4,16 +4,6 @@ module Octokit
   class Client
     # Methods for the Hooks API
     module Hooks
-      # List all Service Hooks supported by GitHub
-      #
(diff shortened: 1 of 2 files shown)


timrogers (approved):

nickfloyd (approved):
🎉

PR #1500, opened 2022-11-02
Outdated credential removal

kfcampbell:
This removes a credential that was unintentionally published, which was discovered and reported through GitHub's Bug Bounty program (https://bounty.github.com).

There is currently no evidence to suggest a compromise of GitHub or any of its systems.


nickfloyd (approved):
💟

PR #1530, opened 2023-01-14
Rename obsolete parameter `IgnoredMethods` to `AllowedMethods`

kyoshidajp:
Fix the following rubocop warning.

> Warning: obsolete parameter `IgnoredMethods` (for `Metrics/BlockLength`) found in .rubocop.yml
`IgnoredMethods` has been renamed to `AllowedMethods` and/or `AllowedPatterns`.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -4,7 +4,7 @@ AllCops:
   TargetRubyVersion: 2.3
   NewCops: enable
 Metrics/BlockLength:
-  IgnoredMethods: ['describe', 'context'] 
+  AllowedMethods: ['describe', 'context'] 
 Layout/BlockAlignment:
   Exclude: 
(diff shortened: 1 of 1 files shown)


kfcampbell (commented):
Thanks for contributing! 

The rename of `Ignored` to `Allowed` is a bit unintuitive to me, but it definitely appears to be the case according to [this code](https://github.com/rubocop/rubocop/blob/master/spec/rubocop/config_obsoletion_spec.rb#L466). 

I would've expected based on [this PR](https://github.com/rubocop/rubocop/pull/9098/files) that `ExcludedMethods` becomes `IgnoredMethods` instead, which would mean that we're in an okay situation right now. 

@kyoshidajp do you mind explaining this to me? Wouldn't the change from `IgnoredMethods` to `AllowedMethods` change the intent of the cop?

nickfloyd (approved):

PR #1533, opened 2023-01-24
Allow using the response_headers or headers when getting Rate Limit Info

### LINKED ISSUE #1400: `Octokit::RateLimit` fields are all `nil` after a `Octokit::TooManyRequests` exception
I'm trying to call `RateLimit#context` (as added in https://github.com/octokit/octokit.rb/pull/1270) but the fields are all `nil`:

```
[2] pry(#<GithubClient>)> e
=> #<Octokit::TooManyRequests: GET https://api.github.com/search/issues?[...]: 403 - API rate limit exceeded for user ID 12345. // See: https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting>
[3] pry(#<GithubClient>)> e.context
=> #<struct Octokit::RateLimit limit=nil, remaining=nil, resets_at=nil, resets_in=nil>
```

Running v4.22.0 (against GitHub Enterprise, in case that matters).

DougEdey:
Resolves #1400 

----

### Before the change?

* `Octokit::TooManyRequests#context.rate_limit` methods return nil

### After the change?

* `Octokit::TooManyRequests#context.rate_limit` methods return the correct values

### Other information

* I've left this as prioritizing the `headers` field over `response_headers`, it's not clear to me when `headers` was valid (I'm not sure if it was an older version of faraday or something), I looked through the code and I think it's because the `on_complete` uses `response_env` (might be [here](https://github.com/octokit/octokit.rb/blob/82765ce829c109644d520d11e86f8673203eb83f/lib/octokit/middleware/follow_redirects.rb#L75) instead of `response` (`Faraday::Response` has `headers` delegating to `env.response_headers`

Note: I can't see where to add in bug fix information in this repo?

----

### Pull request checklist

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

- Bugfix: `Type: Bug`

### DIFF
--- a/spec/octokit/client_spec.rb
+++ b/spec/octokit/client_spec.rb
@@ -1167,12 +1167,13 @@ describe Octokit::Client do
     stub_get('/user').to_return \
       status: 403,
       headers: {
-        'content_type' => 'application/json'
+        'content_type' => 'application/json',
+        'X-RateLimit-Limit' => 60,
+        'X-RateLimit-Remaining' => 42,
+        'X-RateLimit-Reset' => (Time.now + 60).to_i
       },
       body: { message: 'rate limit exceeded' }.to_json
     begin
-      rate_limit_headers = { 'X-RateLimit-Limit' => 60, 'X-RateLimit-Remaining' => 42, 'X-RateLimit-Reset' => (Time.now + 60).to_i }
-      expect_any_instance_of(Faraday::Env).to receive(:headers).at_least(:once).and_return(rate_limit_headers)
       Octokit.get('/user')
     rescue Octokit::TooManyRequests => e
       expect(e.context).to be_an_instance_of(Octokit::RateLimit)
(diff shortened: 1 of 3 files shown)


spec/octokit/client_spec.rb
```diff
@@ -1200,12 +1200,13 @@
     stub_get('/user').to_return \
       status: 403,
       headers: {
-        'content_type' => 'application/json'
+        'content_type' => 'application/json',
+        'X-RateLimit-Limit' => 60,
+        'X-RateLimit-Remaining' => 42,
+        'X-RateLimit-Reset' => (Time.now + 60).to_i
```
DougEdey, line 1206:
```suggestion
        'X-RateLimit-Reset' => (Time.now + 60).to_i
```

nickfloyd (approved):
Hey, @DougEdey Thank you for the contribution here! ❤️ It's g2g!

PR #1256, opened 2020-05-18
Add fixtures server testing

indigok:
With the new generated SDK being built up in #1157, we can move to a different way of testing. All that's needed is testing the generation and a few api endpoints and edge cases. 

Here's a spike of that new testing flow, using the responses from [octokit/fixtures-server](https://github.com/octokit/fixtures-server), for the `repositories` endpoint.

PR #1539, opened 2023-02-03
First try with fixtures

benoittgt:
This is a simple port of https://github.com/octokit/octokit.rb/pull/1256

But while I was doing this first try, it seems that we miss a lot's of fixture to change most of the test ? I do not see a lot of content in https://github.com/octokit/fixtures/tree/main/scenarios/api.github.com

Also I am wondering how much switching to fixtures and having to look at the scenario isn't a little bit too complicated and custom for new comers? 

So before going further. Do we want to keep multiple specs types? Pure mocks with webmocks, VCR tapes and fixtures?


nickfloyd:
Hey @benoittgt,

Oh man, you are hitting on something near and dear to the :octocat: team here in SDKland.  You post such a good observation about developer ergonomics here.

For perspective, we are working on generative prototypes for these SDKs - as in updates to the SDK will happen when the REST API is updated (eventually across all languages) - that's one of the reasons we've been a bit slow to respond.  We're still a bit out from realizing this dream, but that is where we are going.

While that solves this for the future, we still need to solve our problem now.  **I feel at this point, whatever we can do to enhance the dev friendliness of what we currently have while still having the assurance of decent tests is the target we need to be shooting for.**

I will support your choice if you have suggestions or opinions about what feels more approachable. We still want to provide value to folks while we work on the future, so I want to avoid getting railroaded in processes and approaches as blockers.

I don't have any super-strong opinions - I'd love to hear your thoughts.

benoittgt:
Hello @nickfloyd 

The issue I see, it's it is lot's work to offer something that match properly the SDK ecosystem. For me having open api with generative mock is the good idea, but it means that multiple Github packages that rely on Github API should maintain those API schemas together. I think it is out of my capabilities. I use weekly this gem, but cannot invest so much time to make it work now properly and over the time.

We can still migrate some specs to VCR with masked credentials but it means that we need a way to help new users to easily generate personnal token and test to VCR, and we still have the limitation on the repo that can get writable changes (for example editing files). **Or we accept that some change can only be tested via the CI, because this is were the credentials are stored.**

The last proposal is the simpler. I'm happy to try to implement it, and follow PR that could need help on that.

nickfloyd:
@benoittgt agreed. Let's hold off on this for now as we begin to get into the full swing of generative SDKs we might discover a much more simplistic approach to testing and validation. I'll go ahead and close this - thank you for your thoughts and efforts here.

PR #1544, opened 2023-02-14
Specify `TargetRubyVersion: 2.7` in .rubocop.yml

koic:
### Before the change?

Version mismatch between `required_ruby_version` of gemspec and `TargetRubyVersion` of .rubocop.yml. Therefore, some new syntax up to Ruby 2.7 is not used.

### After the change?

gemspec `required_ruby_version` requires 2.7 and higher, so specify `TargetRubyVersion` to 2.7. Some syntax detected by updating to 2.7 applies.

It also removes `Gemspec/RequiredRubyVersion` cop from .rubocop_todo.yml to enable the cop so that version mismatch can be detected.

### Other information

This PR is follow up https://github.com/octokit/octokit.rb/pull/1454.

----

### Pull request checklist

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----

### DIFF
--- a/lib/octokit/client.rb
+++ b/lib/octokit/client.rb
@@ -163,10 +163,10 @@ module Octokit
       inspected.gsub! @bearer_token, '********' if @bearer_token
       # Only show last 4 of token, secret
       if @access_token
-        inspected.gsub! @access_token, "#{'*' * 36}#{@access_token[36..-1]}"
+        inspected.gsub! @access_token, "#{'*' * 36}#{@access_token[36..]}"
       end
       if @client_secret
-        inspected.gsub! @client_secret, "#{'*' * 36}#{@client_secret[36..-1]}"
+        inspected.gsub! @client_secret, "#{'*' * 36}#{@client_secret[36..]}"
       end
 
       inspected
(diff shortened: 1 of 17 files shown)


lib/octokit/client.rb
```diff
@@ -163,10 +163,10 @@ def inspect
       inspected.gsub! @bearer_token, '********' if @bearer_token
       # Only show last 4 of token, secret
       if @access_token
-        inspected.gsub! @access_token, "#{'*' * 36}#{@access_token[36..-1]}"
+        inspected.gsub! @access_token, "#{'*' * 36}#{@access_token[36..]}"
```
kfcampbell, line 166: Do you mind explaining this change to me? Were we only showing the first, second, and third from last characters in the token?

kfcampbell (commented):
I had a quick question about some of the syntax for my own knowledge but this looks great. Thanks!

lib/octokit/client.rb
koic: `String#[s..-1]` and `String#[s..]` are compatible with the endless range syntax added in Ruby 2.6.

kfcampbell (approved):

PR #1566, opened 2023-04-25
Use 'line' instead of 'position'

### LINKED ISSUE #1565: [FEAT]: Add support line property to create_pull_request_comment
### Describe the need

Hi🙋

I want to post a multiline comment using `create_pull_request_comment`, but it seems like I need to use 'line' instead of 'position'. [The API reference](https://docs.github.com/en/rest/pulls/comments?apiVersion=2022-11-28#create-a-comment) mentions that 'position' is deprecated and recommends using 'line' instead. Do you know if there are any plans to support this change? If I missed something, I apologize.

### SDK Version

_No response_

### API Version

2022-11-28

### Relevant log output

_No response_

### Code of Conduct

tick-taku:
Resolves #1565 

----

### Before the change?

* The deprecated 'position' is being used in create_pull_request_comment.(https://docs.github.com/en/rest/pulls/comments?apiVersion=2022-11-28#create-a-comment)

### After the change?

* Use 'line' instead of 'position'.

### Other information

* I changed to specify parameters required for multiline comments, such as 'start_line', in the 'options'.

----

### Pull request checklist
- [x] Added the appropriate label for the given change

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----


kfcampbell:
I've approved the Actions runs; thank you for submitting! I'm slightly torn about this: I think this is a breaking change since the signature of `create_pull_request_comment` is changing. Would it be better to provide an override or mark the current method as deprecated first and then remove it entirely in a new major version after?

tick-taku:
Well...You're right! It may not be possible to provide an override because we're just changing the parameter name, and the type remains the same. So I think we have no choice but to release it in the next major version.
In terms of just using it, you can still use it as is since the type remains the same, and it's not using keyword arguments.
What do you think would be a good course of action?

kfcampbell:
Hmm...I'm not loving any of these options, really. Perhaps the best thing to do is note in comments/docs that "position" will go away in favor of "line" in the next major version, and save this PR up for when we cut a new major version.

tick-taku:
I see, understood! Thanks!
We will put it on hold until the next major version. I will submit a Pull Request with the additional comments.

kfcampbell:
When we merge #1596, we can/should also take this PR before cutting a breaking change.

kfcampbell (approved):
I think the deprecation notice has been baking long enough and we can release this breaking change now!

PR #1569, opened 2023-05-01
Make faraday retry fully optional.

### LINKED ISSUE #1567: [BUG]: The type of the dependency on faraday-retry is irritating and unclear.
### What happened?

When using the gem with `faraday > 2` without having `faraday-retry` installed, `octokit.rb` always prints this warning message.
```
To use retry middleware with Faraday v2.0+, install `faraday-retry` gem
```

According to #1486, there are currently 2 ways to disable the warning:
- disable all warnings by setting `OCTOKIT_SILENT=true`
- install `faraday-retry`

The first option has the side effect of disabling any warning from `octokit.rb`, which is clearly not my intention. The second option requires me to install a gem, but this appears to actually be kind of optional. I had no issues yet without `faraday-retry` and usages test if its constants are defined:
https://github.com/octokit/octokit.rb/blob/fb01ac2add47e5fafe71d25dfb20856cce2a4b46/lib/octokit/default.rb#L36-L42
In case I do not want to use `faraday-retry` for whatever reason, I will therefore always be left with that annoying warning.

From this perspective, it is not clear what kind of dependency `octokit.rb` actually has on `faraday-retry`. It is not required, cannot be listed in the gemspec since `faraday ~> 1.0` is also supported, but also is not fully optional since the exclusion affects all programs using `octokit.rb` by printing a warning.

If it is optional, I would expect to at most get a post-install message of the warning (probably rather a README entry), but not a forced warning. If it is required, I would expect `require 'octokit.rb'` to raise the warning message as an error when `require 'faraday/retry'` raises a `LoadError`.

Implementing either of those options seems quite trivial, contrary to what was claimed in #1486 [in this comment](https://github.com/octokit/octokit.rb/discussions/1486#discussioncomment-3846414). Did I miss something there?

I’ld happily create a PR, although I have no idea which options would be preferred, if any.

### Versions

`octokit.rb = 6.1.1`, `faraday = 2.7.4` without `faraday-retry`

### Relevant log output

_No response_

### Code of Conduct

expeehaa:
Resolves #1567 

----

### Before the change?

When loading the gem with `faraday > 2` and `faraday-retry` not installed, a warning is printed.

### After the change?

When loading the gem with `faraday > 2` and `faraday-retry` not installed, no warning is printed.

### Other information

This is one of the possible solutions for #1567, specifically the one I prefer. I am well aware that the maintainers could prefer another one and, if so, I could modify the pull request accordingly.

----

### Pull request checklist
  - kind of N/A since this has not been tested before

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

- Bugfix: `Type: Bug`

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,9 @@ Access the library in Ruby:
 
     require 'octokit'
 
+Octokit.rb uses the `faraday` gem under the hood. When using `faraday > 2`,
+automatic request retries may be enabled by also installing the gem `faraday-retry`.
+
 ## Making requests
 
 [API methods][] are available as client instance methods.
--- a/lib/octokit/default.rb
+++ b/lib/octokit/default.rb
@@ -10,7 +10,7 @@ if Gem::Version.new(Faraday::VERSION) >= Gem::Version.new('2.0')
   begin
     require 'faraday/retry'
   rescue LoadError
-    Octokit::Warnable.octokit_warn 'To use retry middleware with Faraday v2.0+, install `faraday-retry` gem'
+    # Ignore that faraday-retry is not present.
   end
 end
 


nickfloyd:
Hey @expeehaa,

Thanks for the change here. I'm still trying to work out my thoughts on how we handle this; you make a good point about the noise.  

If you're using Faraday >= 2.x the gem file specifies that `faraday-multipart` and `faraday-retry` should be dependencies

Meaning that you'd have to use the middleware (built-in) for retry if you were using @ >= 1.20 (I belive) and < 2.  Or you would use the separate retry gem if you were >= 2.x - I can see how that implementation is confusing.  

In either case, we should probably require `retry` either through the middleware or with the separate gem - but the support and implementation has to be conditional.  The reason is, we don't know which one is being used - which is why you get the warning. 

I'm am currently at a loss as to how to handle this better. Removing support for Faraday v1 seems like a bit of a drastic change (we cap the version at 3.x) - [this](https://github.com/octokit/octokit.rb/pull/1411) is a bit of a precursor to that. I'm going to chat with @kfcampbell about this as well to see if we can come up with a solution, I just really do not have a good one right now.

----

A note about your change that is orthogonal to our discussion here: 

Specifically, on this change, you'll need to suppress the cop exception by adding:
`- 'lib/octokit/default.rb'` 

To the SupressedException list in this file: `.rubocop_todo.yml`

Or restructure the loading of the gem some other way.

PR #1591, opened 2023-07-07
Add testing for the strings in the actual error message that propagates from Octokit::TooManyRequests

### LINKED ISSUE #1590: [MAINT]: Extend TooManyRequests test to check that the error message still differentiates between primary & secondary rate limits.
### Describe the need

Both primary and secondary API rate limits raise the same `Octokit::TooManyRequests` error. 

I would _prefer_ for 2 separate errors to be raised, but if they're going to be the same, I'd like an additional regex check to make sure the error messages are being transparently passed through in the error.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

arjitj2:
Resolves #1590 

----

### Before the change?
We were just checking that `Octokit::TooManyRequests` was being raised

* 

### After the change?
Now we also check that the error message raised contains a specific regex from Github

* 

### Other information

* 

----

### Pull request checklist

- [x] No

If `Yes`, what's the impact:  

* N/A

### Pull request type

 
 

- Bugfix: `Type: Bug`

----

### DIFF
--- a/spec/octokit/client_spec.rb
+++ b/spec/octokit/client_spec.rb
@@ -849,7 +849,7 @@ describe Octokit::Client do
           content_type: 'application/json'
         },
         body: { message: 'API rate limit exceeded' }.to_json
-      expect { Octokit.get('/users/mojomobo') }.to raise_error Octokit::TooManyRequests
+      expect { Octokit.get('/users/mojomobo') }.to raise_error(Octokit::TooManyRequests, /API rate limit exceeded/)
 
       stub_get('/users/mojomobo').to_return \
         status: 403,
@@ -857,7 +857,7 @@ describe Octokit::Client do
           content_type: 'application/json'
         },
         body: { message: 'You have exceeded a secondary rate limit.' }.to_json
-      expect { Octokit.get('/users/mojomobo') }.to raise_error Octokit::TooManyRequests
+      expect { Octokit.get('/users/mojomobo') }.to raise_error(Octokit::TooManyRequests, /You have exceeded a secondary rate limit./)
 
       stub_get('/user').to_return \
         status: 403,


nickfloyd:
Hey @arjitj2,

Thanks for being part of the community here!  I definitely get where you were headed here, but I don't think the change adds value because:
1. The stub's body covers us on helping us understand why `Octokit::TooManyRequests` was assigned to the result 
2. The match happens in the error class when the 403 is thrown, and the body is evaluated. 

We're trying to test if the error class returns the proper error type and nothing else.  In effect, we are already verifying the match [here](https://github.com/arjitj2/octokit.rb/blob/e3663e7f907715e99af3b72e71efb62cb9c72085/lib/octokit/error.rb), and the passing test affirms that.

Let me know if I missed your intent or if you have any other discussion points on this, I'd be glad to chat about it - but right now it does not appear to add value as far as structural code goes.  Again, thank you for taking the time and working to make things better for the community.

arjitj2:
Hi @nickfloyd , thanks for your response. The goal of this PR is to make sure that Octokit does not hide the error message that led to the `Octokit::TooManyRequests`, and to verify that the users of Octokit can differentiate between the two conditions.

The stub in the test stubs the response from Github, which makes sense. 
```
      stub_get('/users/mojomobo').to_return \
        status: 403,
        headers: {
          content_type: 'application/json'
        },
        body: { message: 'You have exceeded a secondary rate limit.' }.to_json
```

We are trying to affirm that the error message transparently passes through from Octokit. The added change to the assert verifies this.

Without the assertion, I had no way of knowing whether or not our Octokit usage would be able to differentiate whether we're running into primary or secondary rate limits.

nickfloyd:
@arjitj2 Got it. Thanks for the clarity and the work here ❤️

PR #1620, opened 2023-09-07
[feat]: add methods for app hook deliveries

### LINKED ISSUE #1619: [FEAT]: Add methods for app webhooks
### Describe the need

The [endpoints for app webhooks](https://docs.github.com/en/rest/apps/webhooks?apiVersion=2022-11-28) are not yet implemented. I'd like to be able to programmatically list hook deliveries and trigger re-delivery of any deliveries that failed, so these endpoints would be very helpful.

### SDK Version

_No response_

### API Version

_No response_

### Relevant log output

_No response_

### Code of Conduct

SebRollen:
Resolves #1619

----

### Before the change?

* The endpoints for /app/hooks are not yet implemented

### After the change?

* Implementations for listing app webhooks and retriggering delivery have been implemented

### Pull request checklist

This is my first PR to Octokit and I wasn't able to get a test working using the JWT token, would appreciate some pointers here!

- [ ] Yes
- [x] No

----

### DIFF
--- a/lib/octokit/client/apps.rb
+++ b/lib/octokit/client/apps.rb
@@ -226,7 +226,7 @@ module Octokit
       #
       # @return [Array<Hash>] an array of hook deliveries
       def list_app_hook_deliveries(options = {})
-        paginate("app/hook/deliveries", options) do |data, last_response|
+        paginate('app/hook/deliveries', options) do |data, last_response|
           data.concat last_response.data
         end
       end
--- a/lib/octokit/client/apps.rb
+++ b/lib/octokit/client/apps.rb
@@ -217,6 +217,31 @@ module Octokit
       def delete_installation(installation, options = {})
         boolean_from_response :delete, "app/installations/#{installation}", options
       end
+
+      # Returns a list of webhook deliveries for the webhook configured for a GitHub App.
+      #
+      # @param options [Hash] A customizable set of options
+      #
+      # @see https://docs.github.com/en/rest/apps/webhooks#list-deliveries-for-an-app-webhook
+      #
+      # @return [Array<Hash>] an array of hook deliveries
+      def list_app_hook_deliveries(options = {})
+        paginate("app/hook/deliveries", options) do |data, last_response|
+          data.concat last_response.data
+        end
+      end
+
+      # Redeliver a delivery for the webhook configured for a GitHub App.
+      #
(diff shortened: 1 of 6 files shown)


spec/octokit/client/apps_spec.rb
```diff
@@ -278,6 +278,13 @@
       end
     end # .delete_installation
 
+    describe '.list_app_hook_deliveries' do
+      it 'lists hook deliveries for an app' do
+        response = @jwt_client.list_app_hook_deliveries
+        expect(response).to be_kind_of(Array)
```
kfcampbell, line 284: What do you think about expanding the list of assertions here to include a total count check?

lib/octokit/client/apps.rb
```diff
@@ -217,6 +217,31 @@ def find_installation_repositories_for_user(installation, options = {})
...
+
+      # Redeliver a delivery for the webhook configured for a GitHub App.
+      #
+      # @param delivery_id [Integer] The id of a GitHub App Hook Delivery
+      # @param options [Hash] A customizable set of options
+      #
+      # @see https://developer.github.com/v3/apps/#redeliver-a-delivery-for-an-app-webhook
+      #
+      # @return [Boolean] Success
+      def deliver_app_hook(delivery_id, options = {})
```
kfcampbell, line 242: Do you have thoughts on how this might be tested?

spec/octokit/client/apps_spec.rb
SebRollen: Sure, that sounds good to me. I added an assertion that the returned count should be 30, the default per-page amount.

I've got to admit that I'm still a bit confused about how to successfully record a new VCR cassette, which is needed for the tests. Do I need to provide some of the ENV variables listed in [the table](https://github.com/octokit/octokit.rb#running-and-writing-new-tests) in the README and run the specs with those variables?

lib/octokit/client/apps.rb
SebRollen: Yes, just added in a spec for this endpoint as well. Similar to some of the specs for endpoints the delete resources, the only real thing to test for here is a truthy response

SebRollen:
@kfcampbell, friendly bump on this one, I think I addressed all your comments

nickfloyd (approved):
Thanks @SebRollen for these changes! ❤️

PR #1645, opened 2023-12-09
Fix documentation for oauth_applications [ci skip]

gigDevelopment10:
Documentation for method name in oauth_applications.rb is wrong [here ](https://github.com/octokit/octokit.rb/blob/dcbfa148ea1b41d83aac7bd4a39265ef90aab417/lib/octokit/client/oauth_applications.rb#L68) fixed it.


nickfloyd (approved):
Thanks for the clean up here! ❤️

PR #1651, opened 2023-12-29
feat: Renew access token with refresh access token

gigDevelopment10:
Resolves #I1652

----

### Before the change?

* Missing functionality to renew an access token when github app is authenticated.

### After the change?

* Expired access tokens can be renewed using refresh access token.
* Method added in **lib/octokit/client/users.rb**

### Pull request checklist

- [x] No

----


nickfloyd (approved):
Hey @gigDevelopment10 solid commits here! Thank you for doing this! ❤️

PR #1658, opened 2024-01-16
maint: Added base64 gem explicitly dependency

hsbt:
### Before the change?

octokit.rb warns the following message With Ruby 3.3.0.

```
/Users/hsbt/.local/share/gem/gems/octokit-8.0.0/lib/octokit/client/code_scanning.rb:3: warning: base64 was loaded from the standard library, but is not part of the default gems since Ruby 3.4.0. Add base64 to your Gemfile or gemspec. Also contact author of octokit-8.0.0 to add base64 into its gemspec.
```

This means octokit.rb is not working after Ruby 3.4 release. We should add `base64` as dependency.

### After the change?

Above warning is suppressed.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/octokit.gemspec
+++ b/octokit.gemspec
@@ -5,6 +5,7 @@ $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require 'octokit/version'
 
 Gem::Specification.new do |spec|
+  spec.add_dependency 'base64'
   spec.add_dependency 'faraday', '>= 1', '< 3'
   spec.add_dependency 'sawyer', '~> 0.9'
   spec.authors = ['Wynn Netherland', 'Erik Michaels-Ober', 'Clint Shryock']


nickfloyd (changes requested):
Hey @hsbt Thanks for the contributions here!  Before we can move forward would you mind addressing the lint issue of making sure the deps are in the spec file in alphabetical order.  That way CI can run complete.

Alternatively you can give owner commit access to your fork and I can make the changes.

hsbt:
@nickfloyd done to sort alphabetical order.

xuanxu (approved):
:+1:

octokit.gemspec
```diff
@@ -5,6 +5,7 @@ $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require 'octokit/version'
 
 Gem::Specification.new do |spec|
+  spec.add_dependency 'base64'
```
nickfloyd, line 8:
```suggestion
  spec.add_dependency 'base64' if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('3.3')
```

What are thoughts about approaching it like this? I know that it shouldn't matter but I would like to reduce the surface area of the impact.  Given that pre v3.3 can get`base64` from stdlib without the warn I think this approach should be ok.

Let me know your thoughts when you get the chance!  Thanks.

nickfloyd (commented):
Let me know your thoughts on my suggestion around reducing the scope of impact.

nickfloyd (approved):
I am going to favor getting this in given our 3.4 builds/ci are failing - we can discuss the merits of making it a conditional dependency later. 

Thanks @hsbt for the contributions here ❤️ !

octokit.gemspec
hsbt: It's not working. I'm member of Ruby core team and RubyGems team.

RubyGems couldn't handle condition of Ruby version for dependency. RubyGems only support `required_ruby_version` and `required_rubygems_version` for gemspec, not dependency.

nickfloyd: Thanks for the clarity!  I've merged it as is and will ship it once I get a few more changesets in.

hsbt:
Thanks!

PR #1661, opened 2024-01-18
fix: Corrects the order of caching + authorization middlewares

### LINKED ISSUE #1660: [BUG]: HTTP caching middleware returns cached responses to wrong users
### What happened?

In version 4.22.0 (**and onwards**), `http.authorization 'Bearer', @bearer_token` (which immediately sets the `Authorization` header) was changed to `http.request :authorization, 'Bearer', @bearer_token`, which instead adds the `Faraday::Request::Authorization` to the middleware chain, which sets the `Authorization` header when called.

Since the authorization middleware is added at the point of the request, while the caching middleware is added on Faraday initialization (as suggested in the README), it means that the HTTP caching middleware runs first, and the authorization middleware runs second.

This means that when the caching middleware runs, it does not have the `Authorization` header set, and therefore caches requests and then returns them without having the ability to know that these requests are for different users.

For example, `GET /user/installations?per_page=100` should return different results for different users, but with the HTTP caching enabled, it does not.

I have a PR that fixes the issue here: https://github.com/octokit/octokit.rb/pull/1661/files. Open to other approaches as well.

### Versions

Octokit v8.0.0, Ruby v3.2.2

### Relevant log output

_No response_

### Code of Conduct

DimitriosLisenko:
Resolves #1660

----

### Before the change?
The HTTP caching doesn't have the Authorization header available to it, and therefore caches + returns values across user boundaries.

### After the change?
The HTTP caching middleware happens after then authorization middleware, and therefore has the Authorization header and can cache requests per user.

### Pull request checklist

- [ ] Yes
- [x] No

----


DimitriosLisenko:
Alternatively, could use `Faraday::RackBuilder#insert` to place the authorization middleware first in the list of handlers, which guarantees it will run before the caching middleware. Since the authorization middleware just sets some headers, this might be safer to do than rearranging the caching middleware.

PR #1667, opened 2024-02-08
Drop runtime dependency on base64

Earlopain:
#1658 added base64 as a dependency in order to supress a warning/error on Ruby 3.3/3.4.

There are two places here where it is used, and these can instead be replaced by the `pack` method which the base64 gem just wraps around.

I've instead added it as a development dependency so that tests can continue to make use of it.

Other gems have taken a similar approach:
* https://github.com/rubocop/rubocop/pull/12313
* https://github.com/rack/rack/pull/2110
* https://github.com/elastic/elasticsearch-ruby/pull/2295
* https://github.com/socketry/protocol-http/pull/51
* etc.

----

### Before the change?
`base64` is part of the depenency tree.

### After the change?
`base64` is removed.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,7 @@ group :test do
   install_if -> { RUBY_VERSION >= '2.8' } do
     gem 'rexml', '>= 3.2.4'
   end
+  gem 'base64'
   gem 'json', '>= 2.3.0'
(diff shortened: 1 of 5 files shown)


nickfloyd (approved):
Thanks for the thorough clean up on this. ❤️

PR #1678, opened 2024-04-02
Remove outdated version constraint from README

andyw8:
The current release is v8 so this was very out of date. I think it's better just to remove than to keep updating.


nickfloyd (approved):
Thank you for the clean up here ❤️

PR #1696, opened 2024-05-31
feat: Add Ruby 3.3 as supported one.

simi:
Add latest Ruby 3.3 to CI and README.md. Is it time to remove EOL Rubies as well and push major version? I can submit additional PR if welcomed.

---

_Btw. any idea what's with logo in README.md and why https://octokit.github.io/octokit.rb/ is outdated?_


nickfloyd:
> Btw. any idea what's with logo in README.md and why https://octokit.github.io/octokit.rb/ is outdated?

Hey @simi Thanks for the changes here ❤️ 

As far as the logo goes, it's referencing the docs site which is the old way that we referenced the Gundam Cat.  It needs to be updated.

As far as the docs go, this is unfortunately another relic of the past.  We haven't generated those docs using yard for about 5 years now.  Given we are a small team that is[ moving toward Generative SDKs](https://github.blog/2024-01-03-our-move-to-generated-sdks/) the meta needs of some of our older SDKs have simply been deprioritized so that we can focus our efforts on the future.

Please feel free to tackle the out of date docs problem if you'd like and we'd be glad to get it reviewed.  I've created an up for grabs issue [here](https://github.com/octokit/octokit.rb/issues/1700) if you are interested!

nickfloyd (approved):

PR #1706, opened 2024-07-14
Do not warn even if faraday-retry gem is missing

### LINKED ISSUE #1567: [BUG]: The type of the dependency on faraday-retry is irritating and unclear.

akihikodaki:
It makes little sense to emit a warning saying "to use retry middleware" when the user does not request that. Remove the warning.

Resolves #1567

----

### Before the change?

* It emits a warning saying "to use retry middleware..."

### After the change?

* The warning is gone.

### Pull request checklist

- [ ] Yes
- [X] No

----


lib/octokit/default.rb
```diff
@@ -7,11 +7,6 @@
 require 'octokit/warnable'
 
 if Gem::Version.new(Faraday::VERSION) >= Gem::Version.new('2.0')
-  begin
-    require 'faraday/retry'
-  rescue LoadError
-    Octokit::Warnable.octokit_warn 'To use retry middleware with Faraday v2.0+, install `faraday-retry` gem'
```
pbstriker38, line 13: This warning is the only thing that would even let you know that this gem does retries. There is nothing in the README about this. This would be better as an explicit option to enable retries.

akihikodaki: So do you suggest disabling retries by default?

pbstriker38: That would be the best way to remove the warning. The retry functionality needs to be better documented.

akihikodaki: Added commit 3395ed41c0ee7dbc39856c96599acee181638b24 to remove the retry middleware from the default middleware stack and commit c892eb3ad3e5e32e79c537f94df102947eaa92a2 to provide the retry middleware to insert it into the middleware stack explicitly.

README.md
```diff
@@ -631,6 +631,22 @@ x-ratelimit-reset: "1377205443"
...
+It uses [Faraday Retry gem] for Faraday 2.x. Add the gem to your Gemfile
+
+```ruby
+gem 'faraday-retry'
+```
+
+Next, insert `Octokit::Middleware::Retry` before `Octokit::Response::RaiseError`:
+
+```ruby
+Octokit.middleware.insert Octokit::Response::RaiseError, Octokit::Middleware::Retry
```
pbstriker38, line 647: I would think configuring it on the client would be better
```ruby
Octokit::Client.new(auto_retry: true)
```

and then in the connection sawyer options you can insert the retry middleware if @auto_retry
```ruby
conn_opts[:builder].insert(0, *retry_middleware) if @auto_retry
```

akihikodaki: While having a parameter for `Octokit::Client.new` is more concise, there are a few advantages of explicitly requiring to interact with middleware:
1. A user can specify additional options.
2. It is clear that it interacts with middleware.
3. It is consistent with the suggestion for caching, which follows this section in `README.md`.

So there is some trade-off. What do you prefer?

pbstriker38: What happens if you want one client that retries and another that doesn't. Setting on Octokit.middleware is global

akihikodaki: You can also configure an individual instance; see: https://github.com/octokit/octokit.rb/pull/1706/files#diff-7ee18df1929eee9ac41b3ffa5f60cc4904aeb944ff7be6034721b1a76dc2798dR520

akihikodaki:
Rebased to v10.0.0.

PR #1708, opened 2024-07-23
fix: Addresses URI parsing compatibility with Ruby 3.4

casperisfine:
But also older Rubies if the default URI parser is set to `RFC3986`.

### DIFF
--- a/lib/octokit/repository.rb
+++ b/lib/octokit/repository.rb
@@ -81,8 +81,14 @@ module Octokit
 
     private
 
+    ABS_URI_REGEXP = if URI.const_defined?(:RFC2396_PARSER) # Ruby 3.4+
+                       URI::RFC2396_PARSER.regexp.fetch(:ABS_URI)
+                     else
+                       URI::RFC2396_Parser.new.regexp.fetch(:ABS_URI)
+                     end
+
     def validate_owner_and_name!(repo)
-      if @owner.include?('/') || @name.include?('/') || !url.match(URI::ABS_URI)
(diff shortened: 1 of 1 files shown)


nickfloyd:
Hey @casperisfine, this change set looks good.  Thanks for doing it! ❤️ 

Would you mind adding some tests? If not I can do that later.  I'll approve either way and thanks again!

Something like:

```
describe '#validate_owner_and_name!' do
    let(:validator) { described_class.new(owner, name, url) }
    let(:repo) { 'octokit_repo' }

    context 'when owner and name are valid and url is absolute' do
      let(:owner) { 'valid_owner' }
      let(:name) { 'valid_name' }
      let(:url) { 'http://octokit.com' }

      it 'does not raise an error' do
        expect { validator.validate_owner_and_name!(repo) }.not_to raise_error
      end
    end

    context 'when owner contains a slash' do
      let(:owner) { 'invalid/owner' }
      let(:name) { 'valid_name' }
      let(:url) { 'http://octokit.com' }

      it 'raises an invalid repository error' do
        expect { validator.validate_owner_and_name!(repo) }.to raise_error(InvalidRepositoryError)
      end
    end
```

nickfloyd (approved):

casperisfine:
It's already covered by tests in `spec/octokit/repository_spec.rb`.

The issue is that the `Gemfile.lock` contains `uri (0.13.0)` so even on Ruby head it's not yes using the broken version of URI.

The test will only fail once these changes are released and updated in the Gemfile: https://github.com/ruby/uri/pull/107

PR #1712, opened 2024-09-03
fix: Remove the Faraday multipart warning and tweak the error message raised at runtime when Faraday Multipart is not installed

### LINKED ISSUE #1701: [BUG]: `faraday-multipart` warning always printed
### What happened?

Since https://github.com/octokit/octokit.rb/pull/1688 (https://github.com/octokit/octokit.rb/commit/cff7c55bffc312f6ee3c9ceedb8b7929f07cef1d) octokit.rb always prints a warning if it can't load `faraday-multipart`. From the warning, it sounds like I don't need `faraday-multipart` unless I upload licenses. It would be great if the warning was only printed if I tried to do that; or that I could silence the warning somehow.

### Versions

octokit.rb v9.0.0 and up

### Relevant log output

```shell
$ bundle exec ruby -roctokit -e 'p Octokit::VERSION'
To use multipart middleware with Faraday v2.0+, install `faraday-multipart` gem; note: this is used by the ManageGHES client for uploading licenses
"9.1.0"
```

### Code of Conduct

Edouard-chin:
Resolves #1701

----

### Before the change?

* A warning message related to the Faraday Multipart middleware is always printed at boot time.
At runtime, a Faraday error is raised and the message of what happened is not clear.

### After the change?

* The warning message is not printed at boot time. A better error and message is raised at runtime.

### Pull request checklist

- [ ] Yes
- [x] No

----

### DIFF
--- a/lib/octokit/default.rb
+++ b/lib/octokit/default.rb
@@ -12,11 +12,6 @@ if Gem::Version.new(Faraday::VERSION) >= Gem::Version.new('2.0')
   rescue LoadError
     Octokit::Warnable.octokit_warn 'To use retry middleware with Faraday v2.0+, install `faraday-retry` gem'
   end
-  begin
-    require 'faraday/multipart'
-  rescue LoadError
-    Octokit::Warnable.octokit_warn 'To use multipart middleware with Faraday v2.0+, install `faraday-multipart` gem; note: this is used by the ManageGHES client for uploading licenses'
-  end
 end
 
 module Octokit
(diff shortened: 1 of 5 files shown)


nickfloyd (approved):
Thank you for making this better! ❤️ We'll need to ensure with future SDKs that might require a form requests from a kv hash make sure to include this check as well.

nickfloyd:
Labeling as bug because of the poor user experience and noise that is getting logged.

PR #1546, opened 2023-02-16
Add Faraday DEFAULT_EXCEPTIONS for when retry middleware is available

DougEdey:
Resolves #1188 

----

### Before the change?

* Octokit would not retry on timeouts

### After the change?

* Octokit does retry on timeouts

### Other information

* It's pretty straightforward, I can see the original issue marks this as a breaking change.

----

### Pull request checklist
- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been reviewed and added / updated if needed (for bug fixes / features)
- [ ] Added the appropriate label for the given change

### Does this introduce a breaking change?

Please see our docs on [breaking changes](https://github.com/octokit/.github/blob/master/community/breaking_changes.md) to help!

- [ ] Yes (Please add the `Type: Breaking change` label)
- [x] No (I don't think so)

If `Yes`, what's the impact:  

* This _could_ be a breaking change since Octokit will now retry on errors it didn't before, but https://github.com/octokit/octokit.rb/pull/1021 wasn't marked as a breaking change,

### Pull request type

 
 

I want to mark this as a bug, but I can understand if it's a feature!

PR #1724, opened 2024-10-24
Update the "debugging" sample code in `README.md`

areitz:
Update the sample code for enabling HTTP debugging in Faraday to work with newer versions of Faraday.

----

### Before the change?

* In testing using ruby 3.2.3, faraday 2.12.0 and faraday-retry 2.2.1, I found that the included sample code failed with this error:

```
uninitialized constant Faraday::Retry (NameError)

  builder.use Faraday::Retry::Middleware, exceptions: Faraday::Request::Retry::DEFAULT_EXCEPTIONS + [Octokit::ServerError] # or Faraday::Request::Retry for Faraday < 2.0
                            ^^^^^^^^^^^^
	from /Users/$USER/.rbenv/versions/3.2.3/lib/ruby/gems/3.2.0/gems/faraday-2.12.0/lib/faraday/rack_builder.rb:74:in `build'
	from /Users/$USER/.rbenv/versions/3.2.3/lib/ruby/gems/3.2.0/gems/faraday-2.12.0/lib/faraday/rack_builder.rb:63:in `initialize'
```

### After the change?

* After making the proposed to point to this constant: https://github.com/lostisland/faraday-retry/blob/main/lib/faraday/retry/middleware.rb#L18 the example debugging code works successfully.

no

- [ ] Yes
- [x] No

----

### DIFF
--- a/README.md
+++ b/README.md
@@ -597,7 +597,7 @@ traffic:
 
 ```ruby
 stack = Faraday::RackBuilder.new do |builder|
-  builder.use Faraday::Retry::Middleware, exceptions: Faraday::Request::Retry::DEFAULT_EXCEPTIONS + [Octokit::ServerError] # or Faraday::Request::Retry for Faraday < 2.0
+  builder.use Faraday::Retry::Middleware, exceptions: Faraday::Retry::Middleware::DEFAULT_EXCEPTIONS + [Octokit::ServerError] # or Faraday::Request::Retry for Faraday < 2.0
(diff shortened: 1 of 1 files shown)


kfcampbell (approved):
Looks like maybe just an incorrect copy/paste when #1546 was implemented. The actual property seems to have always been `Faraday::Retry::Middleware::DEFAULT_EXCEPTIONS` from what I can tell. Thanks!

PR #1791, opened 2025-12-19
Boolean from response should include all successful 2XX response codes

### LINKED ISSUE #1790: [BUG]: Workflow dispatch events are now suddenly returning 200 for success from the API
### What happened?

It seems like this API: https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event is supposed to return `204` for success, but is now returning `200`. This breaks [actions_workflows](https://github.com/octokit/octokit.rb/blob/main/lib/octokit/client/actions_workflows.rb#L42) code because it relies on `boolean_from_response` which does not treat a pure `200` as success.

**EDIT** It seems like the API now returns the ID and hence this is verifiably a breaking change https://github.com/orgs/community/discussions/9752#discussioncomment-15295321

### Versions

I'm running Octokit 4.25.1, but I checked that the code is the same on `main` as well.

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PatBanks:
Resolves #1790

----

### Before the change?
* Workflow dispatch events started returning 200 rather than 204. boolean_from_response does not consider the entire successful response http code range.

### After the change?
* boolean_from_response now considers any 2XX response code as valid

### Pull request checklist

* I do not believe this introduces a breaking change. The output type of the method does not change.

----

### DIFF
--- a/lib/octokit/connection.rb
+++ b/lib/octokit/connection.rb
@@ -9,7 +9,7 @@ module Octokit
 
     # Header keys that can be passed in options hash to {#get},{#head}
     CONVENIENCE_HEADERS = Set.new(%i[accept content_type])
-    SUCCESSFUL_RESPONSES = [200, 201, 202, 204].freeze
+    SUCCESSFUL_RESPONSES = (200..299)
 
     # Make a HTTP GET request
     #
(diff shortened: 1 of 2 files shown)


enescakir:
It returns HTTP 200 with a body, so it's not a boolean anymore. It should return the body as well.

https://github.com/octokit/octokit.rb/pull/1789

PatBanks:
> It returns HTTP 200 with a body, so it's not a boolean anymore. It should return the body as well.
> 
> #1789

I agree, but is that a breaking change to octokit that would require a major bump? I assumed so. I could reduce this down to just adding the 200 🤷  or if that's wrong I can just close this

